PR #65, opened 2013-08-17
DATAMONGO-653 - Support for index operations in GridFsOperations

aparnachaudhary:
Added support for index operations on GridFsTemplate

PR #105, opened 2013-12-11
DATAMONGO-653 - Support for index operations in GridFsOperations.

thomasdarimont:
Introduced GridFsIndexOperations for creating and deleting GridFS indices which can be accessed via GridFsOperations#indexOps().

Original pull request #65.


mp911de:
We decided to not follow up with this pull request as per comment in the [linked ticket](https://jira.spring.io/browse/DATAMONGO-653).

PR #198, opened 2014-07-02
DATAMONGO-937, DATAMONGO-850, DATAMONGO-976 - Support for $text $search.

christophstrobl:
We now support creating text index based on information gathered on domain types.

Using `@TextIndexed` marks properties to be considered for the full text index. Use the `weight` attribute to influence document scoring during search operations.

Please note that using `@TextIndexed` on entity properties forces all properties of any sub document to be considered as part of the text index. Any set weight will in that case be propagated to the siblings taking the most recent weight information into account, which means that a the weight attribute can be overridden for properties in sub documents.

The default language can be done via `@Document(language)` while `@Language` can be used to define the `language_override` field.

Using `TextQuery` and `TextCriteria` allows creation of queries for `$text $search`.

`{ $meta : “textScore” }` can be included using `TextQuery.includeScore()`. As the fieldname used for `textScore` must not be fixed to _score_ one can use the overload taking the fieldname to override the default.

To map `{ $meta : “textScore” }` we  introduced `@TextScore`. Different to `@Transient` the value will be considered when reading documents.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
@@ -385,7 +385,7 @@ public class MappingMongoConverter extends AbstractMongoConverter implements App
 		entity.doWithProperties(new PropertyHandler<MongoPersistentProperty>() {
 			public void doWithPersistentProperty(MongoPersistentProperty prop) {
 
-				if (prop.equals(idProperty)) {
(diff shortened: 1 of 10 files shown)


amorenew:
@christophstrobl I wonder if it is possible to support automatic force textindexing by `@TextIndexed` 
instead of using manual force indexing like below
```
      TextIndexDefinition textIndex = new TextIndexDefinition.TextIndexDefinitionBuilder()
                .onField("departure", 3F)
                .onField("destination", 2F)
                .onField("description")
                .build();

        this.mongoTemplate.indexOps(FlightInformation.class).ensureIndex(textIndex);
```

PR #231, opened 2014-10-07
DATAMONGO-1066 - Improve handling of empty List values.

thomasdarimont:
We now create a new (mutable) ArrayList instead of a HashSet if given sourceValue (a BasicDBList) is empty. This saves a conversion step.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>1.7.0.BUILD-SNAPSHOT</version>
+	<version>1.7.0.DATAMONGO-1066-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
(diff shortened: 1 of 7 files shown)


zhanggn:
private <S extends Object> S read(TypeInformation<S> type, DBObject dbo, ObjectPath path) {
.......
if (typeToUse.isCollectionLike() && dbo instanceof BasicDBList) {
			return (S) readCollectionOrArray(typeToUse, (BasicDBList) dbo, path);
		}
.........
}
in this method, what if my resultset from mongo is changeable, such as: I define a class like Pax{
@Entity private ArrayList<Telephone> phone;}, but mongo documents are changeful, sometimes this field is List<Telephone>, sometimes is A Telephone. When the documents have List<Telephone>, I could mapping document to JAVA OBJECT well. But, when the documents have A Telephone, I could NOT mapping document to JAVA OBJECT, this field in java object will be null.

mp911de:
Closing this PR in favor of #408.

PR #237, opened 2014-11-06
DATAMONGO-1085 - Sort can not use the metamodel classes generated by QueryDSL (reviewed)

yacota:
DATAMONGO-1085 - Sort can not use the metamodel classes generated by QueryDSL.


thomasdarimont:
We fixed this via 173a62b.

PR #244, opened 2014-11-14
DATAMONGO-941 - Add support for Update $min.

christophstrobl:
We now offer dedicated methods for _min_ (`Number`/`Date`) on `Update`.
As `BigInteger` and `BigDecimal` values are converted to `String` before storing in `MongoDB` it is not possible to use for the _min_ operation. Therefore we type check both the given value and the property it should be applied to and raise an error if  it fails.

This PR ist a first draft for discussion and should not be directly merged.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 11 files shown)


mp911de:
Superseded by [DATAMONGO-1404](https://jira.spring.io/browse/DATAMONGO-1404) and #353.

PR #253, opened 2014-11-28
DATAMONGO-1054 - Add support for fast insertion via MongoRepository.insert(..).

thomasdarimont:
Introduced new insert(..) method variants on MongoRepositories that delegates to MongoTemplate.insert(..). This bypasses ID-population, save event generation and version checking and allows for fast insertion of bulk data.


fdomig:
Prior this change, when extending `SimpleMongoRepository<T, ID>` it was possible to use `protected getMongoOperations()` to get hold of the specific `mongoOperations`. Why has that been removed and what is the suggested alternative for that?

Thanks.

olivergierke:
Keep the `MongoOperations` interface handed into the constructor in a field.

fdomig:
Thanks @olivergierke I just figured moments after adding this comment that this would be the obvious solution. My bad. Thanks though. :-)

PR #258, opened 2014-12-18
Remove manual array copy

alex-on-java:
Remove manual array coping by using Arrays.copyOf(values, values.length)


thomasdarimont:
Thanks merged into 1.7.x, 1.6.x, 1.5.x.

PR #259, opened 2014-12-19
fix wrong parentheses

alex-on-java:
fix JSONParseException passed not as a second argument to IllegalStateException, but into String.format()


thomasdarimont:
Thanks for spotting this - I will apply this via #280.

Cheers,
Thomas

PR #261, opened 2014-12-24
DATAMONGO-1127 - Add support for geoNear queries with distance information.

thomasdarimont:
We now support geoNear queries in Aggregations. Exposed GeoNearOperation factory method in Aggregation. Introduced new distanceField property to NearQuery since it is required for geoNear queries in Aggregations.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
@@ -27,6 +27,7 @@ import org.springframework.data.mongodb.core.aggregation.ExposedFields.ExposedFi
 import org.springframework.data.mongodb.core.aggregation.ExposedFields.FieldReference;
 import org.springframework.data.mongodb.core.aggregation.Fields.AggregationField;
 import org.springframework.data.mongodb.core.query.Criteria;
+import org.springframework.data.mongodb.core.query.NearQuery;
 import org.springframework.data.mongodb.core.query.SerializationUtils;
 import org.springframework.util.Assert;
 
@@ -289,6 +290,17 @@ public class Aggregation {
 	 */
 	public static Fields bind(String name, String target) {
 		return Fields.from(field(name, target));
+	}
+
+	/**
+	 * Creates a new {@link GeoNearOperation} instance from the given {@link NearQuery}.
+	 * 
+	 * @param query
+	 * @return
+	 * @since 1.7
+	 */
+	public static GeoNearOperation geoNear(NearQuery query) {
+		return new GeoNearOperation(query);
 	}
 
 	/**
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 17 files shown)


olivergierke:
I don't quite like the addition of `distanceField` to `NearQuery`, the reason being that it's not applicable to general usage of `NearQuery`. As it basically describes which field the distance should be returned in *in aggregation scenarios*, what do you think of rather adding it to the `GeoNearOperations` type (either as constructor parameter if it's required or as additional builder method).

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
```diff
@@ -292,6 +293,19 @@ public static Fields bind(String name, String target) {
...
 	/**
+	 * Creates a new {@link GeoNearOperation} instance from the given {@link NearQuery} and the distanceField. The
+	 * {@code distanceField} defines output field that contains the calculated distance.
+	 * 
+	 * @param query must not be {@literal null}.
+	 * @param distanceField must not be {@literal null} or empty.
+	 * @return
+	 * @since 1.7
+	 */
+	public static GeoNearOperation geoNear(NearQuery query, String distanceField) {
```
olivergierke: Should we actually add an overload without the distance field applied so that basically people can mimic what they got when using `new GeoNearOperation(nearQuery)` before?

thomasdarimont: the geoNear(...) method was not exposed before this PR.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/GeoNearOperation.java
```diff
@@ -28,11 +28,15 @@
 public class GeoNearOperation implements AggregationOperation {
 
 	private final NearQuery nearQuery;
+	private final String distanceField;
 
-	public GeoNearOperation(NearQuery nearQuery) {
+	public GeoNearOperation(NearQuery nearQuery, String distanceField) {
```
olivergierke: I think we have to overload the constructor to not break existing clients using the old constructor. I suggest to add a hint to favor `Aggregation.geoNear(…)` over using the constructors directly. Should we actually do that for all of the operation constructors that actually have a static factory method equivalent in `Aggregation`?

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/aggregation/AggregationTests.java
```diff
@@ -1018,6 +1021,34 @@ public void shouldRetrieveDateTimeFragementsCorrectly() throws Exception {
 		assertThat(dbo.get("dayOfYearPlus1DayManually"), is((Object) dateTime.plusDays(1).getDayOfYear()));
 	}
 
+	/**
+	 * @see DATAMONGO-1127
+	 */
+	@Test
+	public void shouldSupportGeoNearQueriesForAggregation() {
+
+		mongoTemplate.dropCollection(Venue.class);
```
olivergierke: We should handle these cleanup operations in JUnit callback methods (the same for the drop of the collection at the very end) to make sure they get applied even if exceptions occur.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
olivergierke: I see that. Still, we need to keep the constructor in `GeoNearOperations` which then raises the question why `geoNear(query, distance)` is available as factory method but `geoNear(query)` is not. We should stay symmetric here.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/aggregation/AggregationTests.java
thomasdarimont: ack

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/GeoNearOperation.java
thomasdarimont: You're right. Regarding the hint yes thats a good idea I'll add that.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
thomasdarimont: Ack.

PR #262, opened 2015-01-06
DATAMONGO-1132 - The sample does not match the logical result in the MongoDB repositories section of the documentation

christophe-f:


olivergierke:
Thanks for that! :)

PR #264, opened 2015-01-16
DATAMONGO-1135 - Convert GeoJson to $geometry.

christophstrobl:
We now support [GeoJson](http://www.geojson.org) structures for queries. Depending on the `GeoSpatialIndexType` we use the legacy format for _2D_ and _GeoJson_ for _2D_Sphere_ indexes. Therfore we inspect the query for specific geo spatial keywords and convert the query parameter according to the information we read from the Entities meta model.


christophstrobl:
superseded by #271

PR #269, opened 2015-01-21
Updates to work with the 3.0.x branch of the Java driver

jyemin:
There are a few failing tests, but this is pretty close to working.  Please see individual commits for specific issues.  

The biggest remaining problem relates to checking the results of a write.  From what I can see MongoTemplate currently forces the acknowledgement of every write, regardless of the write concern in play.  This is due to the explicit call by MongoTemplate to WriteResult.getError() (which have been deprecated in 2.13 and removed in 3.0) after each write, and also by the call to WriteResult.getN() after each update.

Given that current behavior, one option is to instead force acknowledgment by using WriteConcern.ACKNOWLEDGED as the minimum write concern.  Doing this lets the driver apply the write concern at the same time as the write, which is the only option you have in 3.0 and more efficient in 2.x as well.

I realize now that I didn't fully follow the contributing guidelines.  I will fix that and rebase if this looks like something that you'll eventually accept.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,8 +30,8 @@
 		<project.type>multi</project.type>
 		<dist.id>spring-data-mongodb</dist.id>
 		<springdata.commons>1.10.0.BUILD-SNAPSHOT</springdata.commons>
-		<mongo>2.12.4</mongo>
-		<mongo.osgi>2.12.4</mongo.osgi>
+		<mongo>3.0.0-SNAPSHOT</mongo>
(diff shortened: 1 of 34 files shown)


jyemin:
Also wanted to note that this branch has some test failures that will be resolved once the next 3.0.0-SNAPSHOT is pushed to Sonatype.  Just need to get past some build failures on our CI server, but hopefully by the end of the day it will be up.

christophstrobl:
Thanks @jyemin we've opend [DATAMONGO-1158](https://jira.spring.io/browse/DATAMONGO-1158) for this one and are actively looking into it.

PR #272, opened 2015-02-05
DATAMONGO-979 - Add support for $size expression in project and group aggregation pipeline.

thomasdarimont:
Introduced AggregationExpression interface to be able to represent arbitrarily MongoDB expressions that can be used in Projection and Group operations.
Supported function expressions are provided via the AggregationFunctionExpressions enum.

Original pull request: #272.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>1.7.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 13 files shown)


olivergierke:
Looks decent to me. On thing I'm not quite sure about is the `String`-based usage of `expression(…)`. Doesn't exposing this makes it easy to shoot yourself into the foot due to typos? I see you added a `size()` method explicitly. Should we simply do this for all expressions available or at least create enums for the supported ones?

thomasdarimont:
The idea with the string based expressions is that users can easily use new features in mongodb without having to wait for us for explicit support.
So we can then introduce dedicated methods and that provide a more typesafe API later.

PR #273, opened 2015-02-12
DATAMONGO-1158 - Add Support for mongo-java-driver 3.0 

christophstrobl:
We now support both mongo-java-driver 2.13 and 3.0 along with MongoDB Server 2.6.7 and 3.0.0.

Pleaes note that some of the configurations options might no longer be valid when used with version 3 of the mongo-java-driver. Have a look at the table below so see some of the major differences in using version 2.x or 3.0

```
                      | 2.x                   | 3.0
----------------------+-----------------------+-----------------------------------------------
default WriteConcern  | NONE                  | UNACKNOWLEDGED
----------------------+-----------------------+-----------------------------------------------
option for slaveOk    | available             | ignored
----------------------+-----------------------+-----------------------------------------------
option for autoConnect| available             | ignored
----------------------+-----------------------+-----------------------------------------------
write result checking | available             | defaults WriteConncern to ACKNOWLEGED for EXCEPTION
----------------------+-----------------------+-----------------------------------------------
rest index cache      | available             | throws UnsupportedOperationException
----------------------+-----------------------+-----------------------------------------------
DBRef resolution      | via DBRef.fetch       | via collection.findOne
----------------------+-----------------------+-----------------------------------------------
MapReduce Options     | applied               | ignored
----------------------+-----------------------+-----------------------------------------------
authentication        | via UserCredentials   | via MongoClient
----------------------+-----------------------+-----------------------------------------------
WriteConcernException | not available         | translated to DataIntegretyViolationException
----------------------+-----------------------+-----------------------------------------------
executeInSession      | available             | requestStart/requestDone commands ignored.
----------------------+-----------------------+-----------------------------------------------
index creation        | via createIndex       | via createIndex
----------------------+-----------------------+-----------------------------------------------
```

This PR will actually not set the version for mongo-java-driver to any 3.0-SNAPSHOT but will remain using 2.13.0. Please run `mvn clean install -P mongo-3-next` to build against lastest mongo-java-driver snapshots.

This pull request should supersed #269 (thanks @jyemin kicking this off - has been an invaluable help!)


jyemin:
Is the behavior change illustrated by MongoTemplateTests#throwsExceptionForDuplicateIds going to be an issue for users?  Comparing the default profile (2.x driver) with mongo-3-next, I see that, although the test passes:

* With the 3.x driver a org.springframework.dao.DuplicateKeyException is thrown
* With the 2.x driver a org.springframework.data.mongodb.core.MongoDataIntegrityViolationException is thrown

And if I change the test to do this regardless of driver version:

    template.setWriteConcern(WriteConcern.UNACKNOWLEDGED);
    template.setWriteResultChecking(WriteResultChecking.EXCEPTION);
 
the behavior with the 2.x driver doesn't change, but with the 3.x driver it fails to throw an exception at all.  

Would you consider changing MongoTemplate such that if WriteResultChecking.EXCEPTION is enabled, and WriteConcern.w < 1, it's changed to WriteConcern.ACKNOWLEDGED?

christophstrobl:
Just ran the tests against MongoDB Server 3.0-rc7 with `--storageEngine=wiredTiger` and found that in that case we've some failing tests...

olivergierke:
A few quick observations:

- we should check all the newly introduced reflective invoker types for hideability in more nested packages (potentially even in package protected visibility) as they're basically mitigators between driver versions. Exposing them publicly in a very root package is a bit suboptimal.
- please keep the `Assert.notNull(…)` etc. methods without a static import. Creating an inconsistency with the rest of the codebase doesn't add any value here and decreases readability. Static imports are fine for JUnit assertions as the methods still contain `assert…`.
- should we have a dedicated `MongoClientFactoryBean` instead of adding `MongoClient` support to `MongoFactoryBean`?
- we should check the changes to `SimpleMongoDbFactory` against Spring Boot's Mongo auto-configuration. I can imagine that while we're retaining binary compatibility, an already build Spring Boot 1.2.x might call the constructor taking a `Mongo` instance with a `MongoClient` parameter. That might just work fine, I'd just like to to double check for unexpected glitches in that area.
- do we need to reject `<mongo:options />` in the XML namespace on 3.0?
- it would be cool if we could add a few notes to the documentation in terms of what users might have to adapt or expect to change if they run on a 3.0 driver

Other than that: a great thank you to both @jyemin and @christophstrobl, looks like quite a way but a great achievement in the end :).

christophstrobl:
Thanks @olivergierke. Already changed package assignment of reflective invokers, static imports and documentation. 
Concerning the changes in `SimpleMongoDbFactory` I altered `spring-data-examples` for mongodb to use latest snapshots, which seemed to work just as expected.

I was not sure about a dedicated `MongoClientFactoryBean` and did not want to add antother piece to this PR. Might be a good idea to potentially do it in an sperate issue.Same applies to `<mongo:options />` where parts are still valid even for the new driver, so I do not think that we should reject it.

jyemin:
@christophstrobl let me know if you want a second set of eyes for those failing tests.

olivergierke:
I've squashed the commits, polished things a little and [published a snapshot of the Mongo 3 specific docs](http://docs.spring.io/spring-data/mongodb/docs/1.7.0.BUILD-SNAPSHOT/reference/html/#mongo.mongo-3) we will publish alongside the release. Thanks again for kicking this off!

PR #274, opened 2015-02-16
DATAMONGO-1165 - Add support for Streaming large result lists.

thomasdarimont:
Initial PoC for a MongoDB Cursor backed Iterator that allows fetching of large result sets.


olivergierke:
There are a few aspects we should improve on here:

- I'd like to avoid additions to `MongoTemplate` / `MongoOperations` in the first place. I think the executions should basically call `MongoOperations.execute(…)` with a callback to make sure we get exception translation applied. Also we need to make sure to publish the application events before and after mapping. `MongoTemplate` contains all necessary callbacks, we might just need to make some of them public to reuse them appropriately.
- I think we should focus on `Iterable` as return type, not `Iterator`.
- While a lot of the wording in the introduced code talks about streams there's no dedicated support for a Java 8 `Stream` yet. While I can see that we probably can generically map from `Iterable` to `Stream` throuhg some code in Spring Data Commons, I think we should be consistent in terms of the terms we use to describe scenarios.
- The newly introduced interfaces probably should make it into Spring Data Commons, too.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/AbstractMongoQuery.java
```diff
@@ -415,4 +429,55 @@ private Object deleteAndConvertResult(Query query, MongoEntityMetadata<?> metada
...
+					return new CloseableIterableCusorAdapter<Object>(cursor, operations.getConverter(), entityType);
+				}
+			});
+		}
+	}
+
+	/**
+	 * @author Thomas Darimont
+	 */
+	final class StreamExecution extends CursorBackedExecution {
```
olivergierke: Do you think it's possible to move that into Spring Data Commons and rather turn it into a post-processing step in case the execution returns a `CloseableIterable` but effectively needs to be a `Stream` (similarly to what we already do with `Optional`s. I think this would allow us to automatically support `Stream` as return type on all stores that implement the execution into a `ClosableIterable`.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/AbstractMongoQuery.java
```diff
@@ -415,4 +429,55 @@ private Object deleteAndConvertResult(Query query, MongoEntityMetadata<?> metada
...
+			return createCursorAndWrapInCloseableIterator(query, javaType, queryMapper);
+		}
+
+		@SuppressWarnings("unchecked")
+		private CloseableIterator<Object> createCursorAndWrapInCloseableIterator(final Query query,
+				final Class<?> entityType, final QueryMapper queryMapper) {
+
+			final MongoPersistentEntity<?> persistentEntity = operations.getMappingContext().getPersistentEntity(entityType);
+
+			return (CloseableIterator<Object>) operations.execute(entityType, new CollectionCallback<Object>() {
```
olivergierke: I am not sure the current setup here works as we expect it. Returning the adapter will basically just port the `DBCursor` to the client in adapter shape. Exceptions thrown while iterating over the result and mapping it still wouldn't be subject to the exception translation implemented in `MongoTemplate`.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoOperations.java
```diff
@@ -959,4 +963,19 @@
 	 * @return
 	 */
 	MongoConverter getConverter();
+
+	/**
+	 * Returns the underlying {@link QueryMapper}.
+	 * 
+	 * @return
+	 */
+	QueryMapper getQueryMapper();
```
olivergierke: Is there any way we can avoid leaking `MongoTemplate` internals here?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/AbstractMongoQuery.java
```diff
@@ -415,4 +420,30 @@ private Object deleteAndConvertResult(Query query, MongoEntityMetadata<?> metada
...
+		}
+	}
+
+	/**
+	 * @author Thomas Darimont
+	 */
+	final class StreamExecution extends CursorBackedExecution {
+
+		@Override
+		Object execute(Query query) {
```
thomasdarimont: I think part of this code could be moved to SD Commons.

PR #284, opened 2015-03-20
DATAMONGO-1188 - Add suport for findAndModify in query derivation.

thomasdarimont:
We now support findAndModify operations on derived query methods.


thomasdarimont:
We should also add a way to configure: FindAndModifyOptions options.

christophstrobl:
Thanks @thomasdarimont updated the PR and finally brought it to the _main_ development line. It's been a pleasure working with you (again 😃 )!

PR #285, opened 2015-03-31
DATAMONGO-990 - Add support for SpEL expressions in @Query.

thomasdarimont:
Ported and adapted support for SpEL expressions @Query annotations from Spring Data JPA. StringBasedMongoQuery can now evaluate SpEL fragments in queries with the help of the given EvaluationContextProvider. Introduced EvaluationContextProvider to AbstractMongoQuery. Exposed access to actual parameter values in MongoParameterAccessor.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2010-2012 the original author or authors.
+ * Copyright 2010-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
```diff
@@ -34,18 +34,23 @@
...
 
 /**
  * Factory to create {@link MongoRepository} instances.
  * 
  * @author Oliver Gierke
+ * @author Thomas Darimont
  */
 public class MongoRepositoryFactory extends RepositoryFactorySupport {
 
+	private static final SpelExpressionParser EXPRESSION_PARSER = new SpelExpressionParser();
```
thomasdarimont: We should move this `SpelExpressionParser` to RepositoryFactorySupport - this could then be shared between multiple store modules.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/repository/SampleEvaluationContextExtension.java
```diff
@@ -0,0 +1,124 @@
...
+
+import org.springframework.data.repository.query.spi.EvaluationContextExtension;
+import org.springframework.data.repository.query.spi.EvaluationContextExtensionSupport;
+
+/**
+ * A sample implementation of a custom {@link EvaluationContextExtension}.
+ * 
+ * @author Thomas Darimont
+ */
+public class SampleEvaluationContextExtension extends EvaluationContextExtensionSupport {
```
thomasdarimont: This is copied from the Spring Data JPA tests - we could create a SD commons test jar to share such test helpers.

PR #286, opened 2015-04-02
DATAMONGO-1193 - Prevent unnecessary db lookups when resolving DBRefs on 2.x driver.

christophstrobl:
We now check against the used driver version before requesting `DB` from `MongoDbFactory`. 

Potential improvements on fetch strategy for MongoDB Java Driver 3 will be handled in [DATAMONGO-1194](https://jira.spring.io/browse/DATAMONGO-1194).

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/DefaultDbRefResolver.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/DefaultDbRefResolver.java
@@ -106,7 +106,7 @@ public class DefaultDbRefResolver implements DbRefResolver {
 	 */
 	@Override
 	public DBObject fetch(DBRef dbRef) {
-		return ReflectiveDBRefResolver.fetch(mongoDbFactory.getDb(), dbRef);
+		return ReflectiveDBRefResolver.fetch(mongoDbFactory, dbRef);
 	}
 
 	/**
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>1.8.0.BUILD-SNAPSHOT</version>
+	<version>1.8.0.DATAMONGO-1193-SNAPSHOT</version>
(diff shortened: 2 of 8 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/DefaultDbRefResolver.java
```diff
@@ -106,7 +107,7 @@ public DBRef createDbRef(org.springframework.data.mongodb.core.mapping.DBRef ann
 	 */
 	@Override
 	public DBObject fetch(DBRef dbRef) {
-		return ReflectiveDBRefResolver.fetch(mongoDbFactory.getDb(), dbRef);
+		return ReflectiveDBRefResolver.fetch(MongoClientVersion.isMongo3Driver() ? mongoDbFactory.getDb() : null, dbRef);
```
olivergierke: Wouldn't the test case become much simple if we just handed the factory into the fetch method? `ReflectiveDbRefResolver` does an `isMongo3Driver()` anyway, so why not just lookup the `DB` within that guard? That would be a simple unit test for `ReflectiveDbRefResolver.fetch(MongoDbFactory, DBRef)` then, wouldn't i?

christophstrobl: the `ReflectiveDbRefResolver` is intended to be a only short living construct (should be removed once we fully switch to MongoDB Java driver 3.x). Therefore I wanted to explicitly check the `DbRefResolver.  We'll need to have further checks for DATAMONGO-1194 there as well.
But sure we can also do it the other way - just let me know.

Since we need to have a look at the DB handling in DATAMONGO-1194 anyway I also tried to keep the changes modest.

olivergierke: We need to move the entire logic of the if-clause up into `DefaultDbRefResolver` if we want to drop 2.x support so it's gonna change anyway. So yeah please go ahead and keep everything in `ReflectiveDbRefResolver.fetch(…)`.

PR #289, opened 2015-04-25
Write the type hint _class always after all other fields.

stefanganzer:
FindAndModify writes the type hint _class either before or after all other fields, whereas save writes it always after. Because of that, MongoDBs addToSet/forEach treats two objects as different, even though they only differ in the position of _class. As a consequence, findAndModify/addToSet/forEach adds duplicates to collections.

The proposed change fixes this by always writing the type hint after all fields.

Unfortunately, this works only for collection fields of complex documents, when both save and findAndModify write the type hint. Please see my other pull request for a test that demonstrates the erroneous behaviour in case of simple documents, where the type hint is written by findAndModify.


christophstrobl:
thank you @stefanganzer! I've created [DATAMONGO-1210](https://jira.spring.io/browse/DATAMONGO-1210) to keep track of this one.

mp911de:
We fixed that issue with [DATAMONGO-1210](https://jira.spring.io/browse/DATAMONGO-1210) (see #292) and a follow-up issue [DATAMONGO-1509](https://jira.spring.io/browse/DATAMONGO-1509) (see also #411) hence closing this PR.

PR #291, opened 2015-04-25
Demonstrate erroneous behaviour of findAndModify

stefanganzer:
With commit b7c3e6965342f774ca1f6df2dd3a22869ecc55b5, the default type hinting was removed. Because of that, when updating documents with nested collections of complex documents like List&lt;List&lt;T&gt;&gt; via findAndModify, the type hints are removed from the complex documents. As a consequence, they cannot be read back:
findAndModifyShouldRetainTypeInformationWithinUpdatedTypeOnDocumentWithNestedCollection
findAndModifyShouldRetainTypeInformationWithinUpdatedTypeOnDocumentWithNestedLists

For documents with an embedded document that has collections, this problem was already present in 1.6.0 (or earlier, not tested):
findAndModifyShouldRetainTypeInformationWithinUpdatedTypeOnEmbeddedDocumentWithCollection
findAndModifyShouldRetainTypeInformationWithinUpdatedTypeOnEmbeddedDocumentWithCollection2


christophstrobl:
fixed in #292

PR #317, opened 2015-08-20
Update MongoChangeSetPersister.java

myArea51:
Add checking for debug enabling in the ```getPersistentId``` method. It is not a serious change, but, I think it would be good to have common approach for whole class to ```debug()``` method using, where parameter is a concatenated ```String```.

[Review on Reviewable](https://reviewable.io/reviews/spring-projects/spring-data-mongodb/317)


mp911de:
Thanks @myArea51, merged with 97da43645

PR #323, opened 2015-09-19
DATAMONGO-1291  Persistent entity introspection to obtain a document annotation through AnnotationUtils

oalles:
Persistent entity introspection to obtain a document annotation through AnnotationUtils instead of jdk introspection facilities.

BasicMongoPersistentEntity captures Mongo metadata of domain objects to be persisted to mongodb (such as collection name) needing for that purpose the Document annotation for that type.
Currently, the annotation is obtained through T.class.getAnnotation() method meaning that a reference is obtained if the annotation is present, else returning null.
Using AnnotationUtils.findAnnotation() instead, could easily extend this behavior resulting in more use cases to annotate domain objects.
As findAnnotation() javadoc states, " it finds a single annotation [...]
The algorithm operates as follows:
1. Search for the annotation on the given class and return it if found. 
2. Recursively search through all annotations that the given class declares. 
3. Recursively search through all interfaces that the given class declares. 
4. Recursively search through the superclass hierarchy of the given class. 
"
Being 1 the current behavior.
2. states findAnnotation() metaannotation support, meaning that i could create my custom annotation annotated with Document. And this is an interesting feature. Even more thinking in Spring 4.2 AliasFor annotation.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/BasicMongoPersistentEntity.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/BasicMongoPersistentEntity.java
@@ -26,6 +26,7 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.expression.BeanFactoryAccessor;
 import org.springframework.context.expression.BeanFactoryResolver;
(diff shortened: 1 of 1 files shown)


christophstrobl:
thanks @oalles - added desired behavior via ddc3925659d30be5a41d3e10103d6089b99e2ceb

oalles:
:clap: Thankyou all for your work,

PR #326, opened 2015-09-22
DATAMONGO-1291 - Read document metadata from composed annotations.

christophstrobl:
We now use `AnnotationUtils.findAnnotation` for `@Document` lookup which enables the full power of Spring composable annotations.

Supersedes: #323


olivergierke:
That's merged for Hopper.

PR #327, opened 2015-09-25
DATAMONGO-934: Implementation of bulk operations

ttrelle:
I added basic support for bulk operation introduced w/ MongoDB 2.6, see https://jira.spring.io/browse/DATAMONGO-934


olivergierke:
That's polished and merged.

PR #339, opened 2015-12-17
DATAMONGO-1348 - Convert GeoJson used in NearQuery.

christophstrobl:
We now convert `GeoJsonPoint` used in `NearQuery` into its according format. This also requires to convert any given `min/maxDistance` as well as the `distanceMultiplier` into meters (metric system).

Along the way we fixed an issue where the actual `Query` used along with `NearQuery.query()` was not properly mapped to the domain types properties.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/NearQuery.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/NearQuery.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.mongodb.core.query;
 
+import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.util.Arrays;
 
 import org.bson.Document;
@@ -25,6 +27,8 @@ import org.springframework.data.geo.Metric;
 import org.springframework.data.geo.Metrics;
 import org.springframework.data.geo.Point;
 import org.springframework.lang.Nullable;
+import org.springframework.data.mongodb.core.geo.GeoJson;
+import org.springframework.data.mongodb.core.geo.GeoJsonPoint;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
 
@@ -46,6 +50,8 @@ public final class NearQuery {
 	private boolean spherical;
 	private @Nullable Long num;
 	private @Nullable Long skip;
+
+	private static int PRECISION = 8;
 
 	/**
 	 * Creates a new {@link NearQuery}.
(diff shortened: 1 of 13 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/NearQuery.java
```diff
@@ -420,25 +426,54 @@ public DBObject toDBObject() {
...
+							RoundingMode.HALF_UP).divide(new BigDecimal(1000), PRECISION + 3,
+							RoundingMode.HALF_UP).doubleValue() : metric.getMultiplier());
 		}
 
 		if (num != null) {
 			dbObject.put("num", num);
 		}
 
-		dbObject.put("near", Arrays.asList(point.getX(), point.getY()));
+		if (point instanceof GeoJsonPoint) {
```
olivergierke: That `instance of` check is used twice here and encapsulated in a method below. I think we should reuse the method to express we're referring to the same conceptual thing. I am not sure I understand enough about the issue being solved here but the need to introduce a lot of private methods makes me feel like we want to introduce a value object to express the underlying concepts more explicitly. Those cascades of `BigDecimal` operations feel rather low level compared to what we're trying to achieve. Could this be operations on `Distance` maybe?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/NearQuery.java
```diff
@@ -420,25 +426,54 @@ public DBObject toDBObject() {
...
 		if (minDistance != null) {
-			dbObject.put("minDistance", minDistance.getNormalizedValue());
+			dbObject.put("minDistance", getDistanceValueInRadiantsOrMeters(minDistance));
 		}
 
 		if (metric != null) {
-			dbObject.put("distanceMultiplier", metric.getMultiplier());
+			dbObject.put(
+					"distanceMultiplier",
+					usesMetricSystem() ? BigDecimal.ONE.divide(getMetricSystemNormalizer(metric), PRECISION + 3,
```
olivergierke: Can you elaborate what this is supposed to implement? I think the calculation should be moved into a dedicated method. I'd also like to see a rationale for the formula.

christophstrobl: to achieve a calculation that is accurate to 0.3 meters one needs to use `BigDecimal` with the given precision

mp911de: Actually, why don't we use plain `double` in here? It has more precision out of the box and lower overhead.

christophstrobl: double does not handle the rounding on infinite decimals correctly which causes trouble.

mp911de:
I would propose the following changes to `org.springframework.data.geo.Metrics` and simplify this PR from there on:

* Add a `METERS` as metric
* Extend the documentation for `Metrics` to mention the base scale is radians and explicitly rename the variable `multiplier` to `radius` (not the getter, just the variable).

Future ideas:
* Add conversion methods to `Metrics` enum instances (similar to `TimeUnit.SECONDS.toNanos(…)` et al.) to convert from one metric to another
* Add `FEET` to support the most commonly used metrics

Does this make sense?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/NearQuery.java
mp911de: `Distance` could do the job if extended for dealing with precision. `Distance` uses `double` for its calculations.

christophstrobl: Still for precise calculations we need to rely on `BigDeciamal` no matter what `Distance` uses currently.

mp911de:
That's merged now.

PR #344, opened 2016-02-21
DATAMONGO-1326 - Add support for $lookup to aggregation.

alessiofachechi:
Now A/F MongoDB $lookup operation is supported.

```
Aggregation.lookup("other_collection", "my_field", "other_field", "my_new_field");
```

For this purpose now `ExposedFieldsAggregationOperationContext` foresees an `additional` mode, which allows for exposing new fields *in addition* to the ones exposed at the moment: unlike the group and project operations, the lookup operation in fact adds a new field to the previous field-set.


christophstrobl:
thanks @alessiofachechi .
that one has been merged via 78e99e6df2591432ab0e38562771baebf671e798. We slightly modified some of the changes and added a few more tests.

alessiofachechi:
Thanks to you @christophstrobl and @mp911de, the code looks more pretty now! :)

PR #346, opened 2016-03-08
fixed validation listener to use basically the same code as in Mappin…

puel:
…gMongoConverter

see: http://sterl.org/2016/03/spring-data-mongodb-jsr-validation/
Issue: https://jira.spring.io/browse/DATAMONGO-1393

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingMongoEventListener.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingMongoEventListener.java
@@ -22,6 +22,7 @@ import javax.validation.Validator;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.data.mongodb.core.convert.LazyLoadingProxy;
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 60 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingMongoEventListener.java
```diff
@@ -54,14 +55,15 @@ public ValidatingMongoEventListener(Validator validator) {
...
-		LOG.debug("Validating object: {}", source);
-		Set violations = validator.validate(source);
+		LOG.debug("Validating object: {}", obj);
+		final Object target = obj instanceof LazyLoadingProxy ? ((LazyLoadingProxy) obj).getTarget() : obj;
+		final Set violations = validator.validate(target);
 
 		if (!violations.isEmpty()) {
 
-			LOG.info("During object: {} validation violations found: {}", source, violations);
+			LOG.info("During object: {} validation violations found: {}", target, violations);
```
puel: unclean but didn't changed --> Logging + Throwing an Exception which may cause double logging of the same Exception / Problem.

puel:
Created a new one with test https://github.com/spring-projects/spring-data-mongodb/pull/350

PR #351, opened 2016-03-23
DATAMONGO-1401 - Fix error when updating entity with both GeoJsonPoint and Version property.

christophstrobl:
We now ignore property reference exceptions when resolving field values that have already been mapped. Eg. in case of an already mapped update extracted from an actual domain type instance.


mp911de:
Merged with 9930ec2.

PR #352, opened 2016-03-24
DATAMONGO-1399 - Allow adding hole to GeoJSON Polygon.

christophstrobl:
We now allow creation of `GeoJsonPolygon` having an outer and multiple inner rings.


mp911de:
That's merged with 0dd9048.

PR #357, opened 2016-04-05
DATAMONGO-1407 - updated jira link to point to correct project on jira

akonczak:


mp911de:
Thanks, that's merged with eee6b62.

PR #367, opened 2016-05-24
DATAMONGO-1437 - Preserve non translatable Exception cause when lazily resolving DBRef.

christophstrobl:


mp911de:
That's merged.

PR #369, opened 2016-06-24
DATAMONGO-1453 - Fix GeoJson conversion when coordinates are Integers.

christophstrobl:
We now use `Number` instead of `Double` for reading "coordinates" from GeoJSON representations.

----

Should at least be back ported to `1.9.x` and `1.8.x`.


mp911de:
That's merged, backported to `1.8.x`, `1.9.x` and forward-ported to `2.0` on `issue/DATAMONGO-1176`.

PR #371, opened 2016-06-28
DATAMONGO-1447 - Add support for isolations on Update.

christophstrobl:
We now allow usage of the [$isolated](https://docs.mongodb.com/manual/reference/operator/update/isolated/#definition) update operator via `Update.isolated()`.
In case isolated is set, the query involved in `MongoOperations.updateMulti` will be enhanced by `'$isolated' : 1` in case the isolation level has not already been set explicitly via eg. `new BasicQuery("{'$isolated' : 0}")`.


mp911de:
That's merged now.

PR #381, opened 2016-07-28
DATAMONGO-1454 - Add support for exists projection in repository query methods.

mp911de:
We now support exists projections for query methods in query methods for derived and string queries.

``` java
public PersonReposotiry extends Repository<Person, String> {

  boolean existsByFirstname(String firstname);

  @Query(value = "{ 'lastname' : ?0 }", exists = true)
  boolean someExistQuery(String lastname);
}
```

---

Related ticket: [DATAMONGO-1454](https://jira.spring.io/browse/DATAMONGO-1454)
Depends on: https://github.com/spring-projects/spring-data-commons/pull/171

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>1.10.0.BUILD-SNAPSHOT</version>
+	<version>1.10.0.DATAMONGO-1454-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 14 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/Query.java
```diff
@@ -63,6 +63,14 @@
 	boolean count() default false;
 
 	/**
+	 * Returns whether the query defined should be executed as exists projection.
+	 *
+	 * @since 1.10
+	 * @return
+	 */
+	boolean exists() default false;
```
olivergierke: I wonder if we could derive that from the return type of the query method instead. Queries on MongoDB *always* return a document which can't be mapped to a `boolean` naturally. So we could just enable the exists execution for all query methods that return a `boolean`, can't we?

mp911de: That works for string-based queries quite well. I added a commit to derive count queries from a `long` return type.

PR #389, opened 2016-09-13
DATAMONGO-1490 - Change the XML data type of boolean flags to String.

mp911de:
We now accept String data types for boolean flags in XML configurations. Boolean data types in the XSD don't allows use of property placeholders even if the resolved value could be converted to boolean. Affected fields by this change are:

* `<mongo:repositories create-query-indexes=… />`
* `<mongo:options ssl=…/>`
* `<mongo:client-options ssl=… />`

----

Related ticket: [DATAMONGO-1490](https://jira.spring.io/browse/DATAMONGO-1490)


spring-data-mongodb/src/main/resources/org/springframework/data/mongodb/config/spring-mongo-1.8.xsd
```diff
@@ -37,8 +37,8 @@ Defines a MongoClient instance used for accessing MongoDB.
 				<tool:annotation>
```
mp911de: Let's remove changes to this file entirely from the PR.

PR #395, opened 2016-10-07
Correct query keyword 'Between', now also contains the range values.

hoppermann:
DATAMONGO-1506 - Different behavior in the query keyword 'Between' between the MongoDB and the JPA implementation

Fix for [DATAMONGO-1506](https://jira.spring.io/browse/DATAMONGO-1506)


mp911de:
Thanks for the PR. We currently don't plan to change the behavior. Closing this PR according to comments in [DATAMONGO-1506](https://jira.spring.io/browse/DATAMONGO-1506).

PR #398, opened 2016-10-09
Adds missing @Test annotation

LinkedList:
I don't think Jira issue is necessary for this.


mp911de:
Thanks for your PR. We create Jira tickets for each change to keep track of the context. That's merged and ported to the related branches.

PR #399, opened 2016-10-10
DATAMONGO-1508 - documentation for db-factory authentication-dbname a…

wheezil:
It was difficult to figure out how to use a separate authentication db from the target db.  So I added an example of that to the documentation.  Please be gentle, it is my first pull request ever ;-)
John Lilley


mp911de:
Thanks for your PR. That's merged, polished and ported to the related branches.

wheezil:
Yay!  Thanks for the update.

PR #403, opened 2016-11-02
DATAMONGO-1520 - Aggregation.match should accept CriteriaDefinition.

pavelvodrazka:
Use the same argument type for Aggregation.match method and MatchOperation constructor. Please revise.


christophstrobl:
thank you for proposing this change. Unfortunately it breaks binary compatibility. We added `match(CriteriaDefinition)` next to the existing.

PR #408, opened 2016-11-15
DATAMONGO-1525 Converting empty EnumSet

faissalwise:
DATAMONGO-1525 Converting empty EnumSet


mp911de:
Thanks for the PR. Can you confirm signing the [CLA](https://cla.pivotal.io/sign/spring/icla)? If so, then I'd take the PR from here.

olivergierke:
Already fixed in both master and 2.0.x, 1.9.x and 1.8.x.

PR #409, opened 2016-11-16
DATAMONGO-1506 - Different behavior in the query keyword 'Between' be…

faissalwise:
DATAMONGO-1506 -Different behavior in the query keyword 'Between' between the MongoDB and the JPA implementation


mp911de:
Thanks for the PR. We currently don't plan to change the behavior. Closing this PR according to comments in [DATAMONGO-1506](https://jira.spring.io/browse/DATAMONGO-1506).

PR #410, opened 2016-11-16
DATAMONGO-1530: support missing pipeline operators available in Mongo…

sebge2emasphere:
Pull request for https://jira.spring.io/browse/DATAMONGO-1530

PR #413, opened 2016-11-23
DATAMONGO-1530 - Add support for missing aggregation operators in SpEL based support.

christophstrobl:
We now allow usage of mostly all aggregation operation expressions via SpEL on pipeline stage.

Supersedes (and contains changes from): #410


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/spel/MethodReferenceNode.java
```diff
@@ -63,13 +146,42 @@
...
 		return FUNCTIONS.get(methodName);
 	}
+
+	/**
+	 * @author Christoph Strobl
+	 * @since 1.10
+	 */
+	@lombok.Data
+	@lombok.Builder
+	public static class AggregationMethodReference {
```
mp911de: How about an custom written builder that is not `public` and is tailored to the construction flow (i.e. method name always required, adoption of argument type characteristics)?

PR #414, opened 2016-11-23
DATAMONGO-784 - Add support for comparison aggregation operators to $group & $project.

christophstrobl:
We now directly support comparison aggregation operators (`$cmp`, `$eq`, `$gt`, `$gte`, `$lt`, `$lte` and `$ne`) on both `$group` and `$project` stages.


mp911de:
That's merged to `master` and forward-ported to `2.0.x`.

PR #415, opened 2016-11-24
DATAMONGO-1534 - Fix bulk operations insert missing to write type info.

christophstrobl:
We now correctly convert entities into their MongoDB representation including type information via `_class` property when using bulk insert.

---

should be back ported to `1.9.x`.


mp911de:
That's merged to `master`, `1.9.x` and `2.0.x`.

PR #417, opened 2016-11-25
DATAMONGO-1538 - Add support for $let to $project aggregation.

christophstrobl:
We now support `$let` in aggregation `$project` stage.

```
newAggregation(Sales.class,
	project()
		.and(let(ADD.of(field("price"), field("tax"))).as("total")
			.andExpression(
				ConditionalOperator.newBuilder().when(field("applyDiscount")).then(0.9D).otherwise(1.0D))
			.as("discounted")
			.in(MULTIPLY.of(field("total"), field("discounted"))))
		.as("finalTotal"));
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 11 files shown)


mp911de:
That's merged.

PR #420, opened 2016-12-05
DATAMONGO-1540 - Add support for $map (aggregation).

christophstrobl:
We now support the `$map` operator in aggregation.


mp911de:
That's merged.

PR #425, opened 2016-12-08
Documentation fixes for invalid class MongoMappingConverter

kdombeck:
There are several references to `MongoMappingConverter` which is not a valid class. Instead it should be referring to `MappingMongoConverter `. Looking at the history of the documentation it looks like it has been this way for a very long time.

Also cleaned up a couple of JavaDoc and misspelling issues in `MappingMongoConverter`.


mp911de:
Thanks for your PR. That's squashed and merged now.

PR #431, opened 2016-12-19
DATAMONGO-1467 - Add support for MongoDB 3.2 partialFilterExpression for index creation.

christophstrobl:
We now support `partialFilterExpression` via `partial`. This allows to create partial indexes that only index the documents in a collection that meet a specified filter expression.

```java
new Index().named("idx").on("k3y", ASC).partial(filter(where("age").gte(10)))
```
The filter expression can be set via a plain `DBObject` or a `CriteriaDefinition` and is mapped against the associated domain type.

---

Relates to: #380

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>1.10.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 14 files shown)


vibhaG:
Does the Mongo compound index support partial ind.ex filters? 
For example {ID: 1, realm: 1},
    {unique: true, 
        partialFilterExpression: { type: { $eq: "externalUser" } } }
)
If it does, I was not able to figure out how that works?

PR #435, opened 2017-01-13
DATAMONGO-1588 - Fix derived finder not accepting subclass of parameter type.

christophstrobl:
We now allow using sub types as arguments for derived queries. This makes it possible to use eg. a `GeoJsonPoint` for querying while the declared property type in the domain object remains a regular (legacy) `Point`.


mp911de:
That's merged.

PR #438, opened 2017-02-03
DATAMONGO-1607 - Fixed Double casts for Circle, Point and Sphere.

thiagodsti:


christophstrobl:
Thanks @thiagodsti! Applied to master and ported to SR branches.

PR #441, opened 2017-02-09
DATAMONGO-1603 - Fix Placeholder not replaced correctly in @Query.

christophstrobl:
Fix issues when placeholders are appended with other chars eg. `'?0xyz'` or have been reused multiple times within the query.


mp911de:
That's merged.

PR #442, opened 2017-02-18
#DATAMONGO-1561

yacota:
Full text indexing nested documents when there are concrete fields to be indexed is a little bit overkill IMO. 
With current behaviour when a new field is added to the nested entity an exception is thrown because text index's definition can not change(only one text index per collection is allowed), this PR tries to avoid this problem


mp911de:
The repo underwent since this PR was created a few significant changes so that this PR cannot be merged anymore. Sorry for the inconvenience.

PR #448, opened 2017-03-06
DATAMONGO-1421 - Fix serialization in error message causing error itself.

christophstrobl:
We now make sure to safely serialize the criteria object used for creating the error message when raising an `InvalidMongoDbApiUsageException` in cases where `addCriteria` is used to add multiple entries for the same property.


mp911de:
That's merged.

PR #449, opened 2017-03-08
DATAMONGO-1620 - Add server-selection-timeout to XML MongoClientOptions config.

christophstrobl:
This change allows in combination with a MongoDB Java Driver `3.x` setting the `serverSelectionTimeout` on `MongoClientOptions` via the  `server-selection-timeout` XML attribute.

The attribute is ignored when set using a `2.x` MongoDB driver. 

This change should be ported to  `1.10.x` (the namespace xsd (1.10.2) is already set up for it).
For `2.x` the `server-selection-timeout` should be added leaving out the reflection fluff required in `MongoClientOptionsFactoryBean` for`1.x`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>1.11.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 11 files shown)


mp911de:
That's merged, polished and ported to the according branches. Removed reflection access for `2.0.x`.

PR #450, opened 2017-03-17
DATAMONGO-1645 - Safely serialize JSON output for log message in LoggingEventListener.

christophstrobl:
We now make sure to safely serialize JSON output for mapped documents. This prevents the logger from rendering false exception messages to log appender.


mp911de:
That's merged, polished and ported to the according branches.

PR #451, opened 2017-03-17
DATAMONGO-1643 - Add 2.x namespace xsd.

christophstrobl:
Add 2.x namespace xsd.

Remove and replace usage of `mongo` by `mongoClient`. This involves xsd schema, bean names, constructor and parameter types. This required some API changes as some server commands are no longer directly available through the api, but have to be invoked via `runCommand`.

Also remove references to outdated API using Credentials and an authentication DB instead of `MongoCredentials` for authentication.

Updated and removed (now unused) tests; Altered documentation.


mp911de:
That's merged now.

PR #457, opened 2017-04-12
DATAMONGO-1666 - Consider collection type in bulk DBRef fetching.

mp911de:
We now consider the property's collection type after bulk-fetching DBRefs before returning the actual result value. The issue got only visible if bulk fetching is possible and constructor creation is used. Setting the property value on through an property accessor works fine because the property accessor checks all values for assignability and potentially converts values to their target type. That's different for constructor creation.

---

Related ticket: [DATAMONGO-1666](https://jira.spring.io/browse/DATAMONGO-1666).

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
@@ -31,6 +31,7 @@ import org.bson.Document;
 import org.bson.conversions.Bson;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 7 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
```diff
@@ -913,7 +912,10 @@ private Object readCollectionOrArray(TypeInformation<?> targetType, List sourceV
 		}
 
 		if (!DBRef.class.equals(rawComponentType) && isCollectionOfDbRefWhereBulkFetchIsPossible(sourceValue)) {
-			return bulkReadAndConvertDBRefs((List<DBRef>) (List) (sourceValue), componentType, path, rawComponentType);
```
olivergierke: That's a funny one. I didn't even know casting something twice was allowed :D.

olivergierke:
Looks good to me.

PR #459, opened 2017-05-02
DATAMONGO-1518 - Add support for Collation.

christophstrobl:
We now support ``Collation``s for collections, indexes, queries, findAndModify, delete, geo, bulk and aggregation operations in both the imperative and reactive implementations on template level.

We still need to think about collation support via annotations for the repository layer and QueryByExampleExecuter.


mp911de:
That's merged now.

PR #463, opened 2017-05-15
DATAMONGO-1689 Add Kotlin extensions for [Reactive]MongoOperations.

sdeleuze:


spring-data-mongodb/pom.xml
```diff
@@ -230,13 +231,93 @@
 			<artifactId>spring-webmvc</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
olivergierke: Should these be optional dependencies?

spring-data-mongodb/pom.xml
```diff
@@ -230,13 +231,93 @@
...
+							</sourceDirs>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<executions>
```
olivergierke: Can you elaborate on the need for this quite significant change in the compile settings?

spring-data-mongodb/pom.xml
sdeleuze: Indeed, I will change that.

spring-data-mongodb/pom.xml
sdeleuze: It seems to be the [recommanded way to handle Kotlin + Java Maven builds](https://kotlinlang.org/docs/reference/using-maven.html#compiling-kotlin-and-java-sources).

JetBrains explantation is the following:
> To compile mixed code applications Kotlin compiler should be invoked before Java compiler. In maven terms that means kotlin-maven-plugin should be run before maven-compiler-plugin using the following method, making sure that the kotlin plugin is above the maven-compiler-plugin in your pom.xml file.

spring-data-mongodb/pom.xml
sdeleuze: Updated

sdeleuze:
Another thing we could do is to provide extensions with alternative function names to `in`  and `is` which are Kotlin reserved keywords and [currently need to be escaped](https://github.com/mixitconf/mixit/blob/master/src/main/kotlin/mixit/repository/UserRepository.kt#L43).

We should also update the `@since`  version which was set by mistake to `5.0` instead of `2.0`.

sdeleuze:
I will send an updated version of this PR tomorrow with the following changes:
 - Fluent alternative for `MongoOperations` extensions
 - `@since 2.0`
 - Removal of `KClass` based variants and keep only reified type parameter variant as explained on [SPR-15660|https://jira.spring.io/browse/SPR-15660]

Could we consider inclusion in master when I will have pushed this updated PR?

christophstrobl:
@sdeleuze I already started working on integrating Kotlin in this branch (https://github.com/spring-projects/spring-data-mongodb/tree/issue/DATAMONGO-1689). 
I also extracted the plugin part (see spring-projects/spring-data-build#370).

sdeleuze:
Ok thanks I will send the update against your branch and will answer on the spring-data-build side.

olivergierke:
Given the short amount of time before the M4 release tomorrow, I'd love to keep the build changes in this module for now.

@sdeleuze do you think you can get your changes in tonight or early tomorrow morning CET?

PR #466, opened 2017-06-02
DATAMONGO-1563 - Add fluent alternative for MongoOperations.

christophstrobl:
We now provide an alternative API for `MongoOperations` that allows defining operations in a more fluent way. This reduces the number of methods and strips down the interface to a minimum while offering a more readable API.

```java
// find all with filter query and projecting return type
template.query(Person.class)
    .matching(query(where("firstname").is("luke")))
    .as(Jedi.class)
    .all();

// update with filter & upsert 
template.update(Person.class)
    .apply(new Update().set("firstname", "Han"))
    .matching(query(where("id").is("han-solo")))
    .upsert();

// remove all matching
template.remove(Jedi.class)
    .inCollection("star-wars")
    .matching(query(where("name").is("luke")))
    .all();

// aggregate
template.aggregateAndReturn(Jedi.class)
    .inCollection("star-wars")
    .by(newAggregation(project("name")))
    .get();
```


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ExecutableAggregationOperationBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+		 */
+		WithAggregationBuilder<T> inCollection(String collection);
+	}
+
+	/**
+	 * Trigger execution by calling one of the terminating methods.
+	 *
+	 * @param <T>
+	 */
+	interface AggregateOperationBuilderTerminatingOperations<T> {
```
mp911de: How about just `AggregationOperations`?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ExecutableAggregationOperationBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+		 */
+		AggregationResults<T> get();
+
+		/**
+		 * Apply pipeline operations as specified. <br />
+		 * Returns a {@link CloseableIterator} that wraps the a Mongo DB {@link com.mongodb.Cursor}
+		 *
+		 * @return
+		 */
+		CloseableIterator<T> stream();
```
mp911de: I'm a bit torn on `CloseableIterator`. `Streamable` doesn't seem to be the right interface here because we expect an executed statement, which is `Iterator`. `Iterator` to Java 8 `Stream` adoption is sort of cumbersome and I wonder whether there might be an interface from our side that is `Closeable` and able to return a `Stream`

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ExecutableFindOperationBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+		/**
+		 * [optional] Define the target type fields should be mapped to. <br />
+		 * Just skip this step if you are anyway only interested in the original domain type.
+		 *
+		 * @param resultType must not be {@literal null}.
+		 * @param <T>
+		 * @return
+		 * @throws IllegalArgumentException if resultType is {@literal null}.
+		 */
+		<T1> WithQueryBuilder<T1> as(Class<T1> resultType);
```
mp911de: Minor: How about `R` instead of `T1`? (R -> Result)

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ExecutableAggregationOperationBuilder.java
christophstrobl: had some back and forth on the names, generally it turned out that we already have similar named components for quite a few of the ones introduced here. So there'd already be `AggregationOperation`. I do see the point that for having eg. an instance var `AggregateOperationBuilderTerminatingOperations findAllInStarWars;` does not read nice compared to `AggregationOperation findAllInStarWars;`. Bottom line, 
I'm really open for better naming as long as we can keep it consistent across the interfaces for this feature.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ExecutableAggregationOperationBuilder.java
christophstrobl: The main goal here was to provide users with an alternative way of interacting with `MongoOperations` therefore I'd not surprise them with different return types for pretty much the same operation.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ExecutableFindOperationBuilder.java
christophstrobl: 👍 , even `@param` in JavaDoc does not match `<T1>`.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ExecutableFindOperationSupport.java
```diff
@@ -0,0 +1,202 @@
...
+		public DelegatingQueryCursorPreparer(CursorPreparer delegate) {
+			this.delegate = delegate;
+		}
+
+		@Override
+		public FindIterable<Document> prepare(FindIterable<Document> cursor) {
+
+			FindIterable<Document> target = delegate.prepare(cursor);
+
+			if (limit.isPresent()) {
```
olivergierke: Rather `return limit.map(it -> target.limit(it)).orElse(target)`?

sdeleuze:
I am very interested in leveraging this API in https://github.com/mixitconf/mixit, let me know when available in master (I can update DATAMONGO-1689 to add Kotlin extensions for this new API).

PR #472, opened 2017-06-23
DATAMONGO-1678 - Run bulk update / remove operations through Query-/UpdateMappers.

christophstrobl:
We now make sure to run any query / update object through the `Query-` / `UpdateMapper`. This ensures `@Field` annotations and potential custom conversions get processed correctly for update / remove operations.


mp911de:
That's merged, polished and back-ported.

PR #473, opened 2017-06-23
Add Kotlin extensions for Criteria API.

sdeleuze:
This commit introduces 2 Criteria method aliases because
`in` and `is` are reserved keywords in Kotlin:
 - `isEqualTo` alias for is
 - `inValues` alias for in

I have considered using `containsAny` instead of `inValues` but I think the later will be more discoverable via autocomplete.


olivergierke:
That's merged, thanks!

PR #474, opened 2017-06-23
	DATAMONGO-1682 - Add support partialFilterExpression for reactive index creation.

christophstrobl:
We now support partial filter expression on indexes via `Index.partial(…)` on the reactive API. This allows to create partial indexes that only index the documents in a collection that meet a specified filter expression.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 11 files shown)


mp911de:
That's merged and polished.

PARASRAUTELA:
@christophstrobl @mp911de  I'm not able to find the usage for this. Can you help with a sample snippet on how to create partialFIlterExpression Index preferably with annotation? Quick googling around didn't help much.
Thanks in advance.

PR #475, opened 2017-06-29
DATAMONGO-1726 - Add oneValue() & findValue() to FluentMongo API returning null instead of Optional.

christophstrobl:
We leave the choice of using `Optional` open by also providing terminating find operation methods that return `null` instead of `Optional`.


olivergierke:
LGT, except the commit message stating `findValue()`. Guess it should've been `firstValue()`.

mp911de:
That's merged now.

PR #484, opened 2017-07-11
DATAMONGO-1738 - Move repository query execution to fluent operations API.

olivergierke:
We now use the fluent FindOperations API in AbstractMongoQuery and MongoQueryExecution instead of the MongoOperations. This allows us to eagerly resolve some general execution coordinates (which collection to query etc.) and thus simplify the eventual execution.

Got rid of a couple of very simple QueryExecution implementations that can be replace by a simple lambda. Removed the need to read into a partially filled domain object and then map to the projection DTO as we can now tell the operations to read into the DTO directly.

Adapted unit tests.

Open points:

- if the fluent operations returned a `Stream` for the call to `….stream()` immediately we could even further reduce complexity of the stream execution

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>2.0.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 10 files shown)


olivergierke:
Needs #485 first.

mp911de:

PR #485, opened 2017-07-11
DATAMONGO-1739 - Changed TerminatingFindOperation.stream() to return Stream.

olivergierke:
TerminatingFindOperation.stream() now returns a Stream directly, leveraging Spring Data Commons' StreamUtils.createStreamFromIterator(…) to create a Stream and register a callback to forward calls to Stream.close() to the iterator.


mp911de:
That's merged now.

PR #491, opened 2017-07-28
DATAMONGO-1756 - Fix nested field name resolution for aggregation.

christophstrobl:


mp911de:

PR #496, opened 2017-08-21
DATAMONGO-1768 - Allow ignoring type restriction when issuing QBE.

christophstrobl:
We now allow to remove the type restriction inferred by the QBE mapping via an _ignored path_ expression on the `ExampleMatcher`. This allows to create sort of untyped QBE expressions returning all entities matching the query without limiting the result to types assignable to the probe itself.

2fa07bb might be a candidate for back porting to 1.x generation, while 6e9abbf depends on changes in spring-projects/spring-data-commons#238 which opens the API so that creation of an `UntypedExample` is possible.


mp911de:
That's merged, polished and back-ported.

PR #497, opened 2017-09-05
DATAMONGO-1772 - Fix UpdateMapper not rendering type key correctly for abstract lists elements contained in concrete typed ones.

christophstrobl:


mp911de:

PR #498, opened 2017-09-05
DATAMONGO-1774 - Fix infinite loop in ReactiveMongoOperations#remove(Mono, String)

christophstrobl:


mp911de:

PR #505, opened 2017-10-17
DATAMONGO-1802 - Convert Binary data on read to byte[] if necessary.

christophstrobl:
Should be back ported to _2.0.x_.


mp911de:

PR #508, opened 2017-10-23
DATAMONGO-1809 - Fix positional parameter detection for PropertyPaths.

christophstrobl:
We now make sure to capture all digits for positional parameters.


mp911de:

PR #514, opened 2017-11-07
DATAMONGO-1761 - Add support for distinct(…).

christophstrobl:
We now support `distinct` throughout the imperative and reactive templates. We also added `distinct` to the fluent API counterpart and provided the required Kotlin extensions for the newly added methods.

This now allows obtaining distinct field values for a single field. The stored values do not have to have the same data type to be considered, nor is the feature limited to simple types.

```java
template.query(Person.class)
  .distinct("lastname")
  .as(String.class)
  .all();
```

It is recommended, when obtaining field values of different types, to use `Object` for the resulting type. Fixing the type might lead to conversion errors for non convertible types.

Fields containing `DBRef` will be resolved and converted to the corresponding domain type.

----

Supersedes and builds upon: #494 (thank you @duozhilin)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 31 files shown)


mp911de:
That's now merged and polished.

PR #515, opened 2017-11-17
DATAMONGO-1824 - Assert compatibility with MongoDB 3.6

christophstrobl:
We now use the driver native option for `aggregate` instead of a plain command execution. This is necessary as of MongoDB 3.6 the cursor option is required for aggregations changing the return and execution model.

Still we maintain the raw values of `AggregationResult` as we used to do before. However, relying on `executeCommand` to change aggregation behavior in custom code will no longer work.

Along the the way we opened up `Aggregation` itself to expose the `AggregationOptions` in use und deprecated the `$pushAll` option on `Update` since it has been removed in MongoDB 3.6.

---

Tested against MongoDB: 3.6.RC3, 3.4.9 and 3.2.6

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
+	<version>2.1.0.DATAMONGO-1824-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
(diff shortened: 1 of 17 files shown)


mp911de:

chengbinTechnology:
But The Project is Compiler by jdk 1.7, Version 2.0+ not supported

chengbinTechnology:
Can you release a version that supports jdk 1. 7

mp911de:
These changes are also shipped with Spring Data MongoDB version 1.10.10 and higher requiring Java 6.

PR #518, opened 2017-11-29
Match a column with multiple values in findAll() method using Query, Example and Pageable

ketankkp27:
Add findAll() method which takes Query, Example and Pageable and return appropriate result.

Currently it supports example and pageable type method but after implement above method it will support like custom query with example, query and pageable  at the same time.

For example:- 
There is table called Person.
name:
dob:
age: etc 20 more fields. 

Expected search result is like dob grater than x year and less than y year and name like RAM or Shyam and want to match some other fields.

In this case developer can not use example class for name and dob. Because Example class can match with only one value. But we want to bring data with multiple condition on same field, like name is Ram or Shyam and DOB is gt x and lt Y.

That's why we have introduced above method which will allow them to write custom query and also allow use of example query class with pageable response which will resolve many problem like above example.

- [ ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAMONGO).

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleMongoRepository.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleMongoRepository.java
@@ -283,6 +283,26 @@ public class SimpleMongoRepository<T, ID> implements MongoRepository<T, ID> {
 		return PageableExecutionUtils.getPage(list, pageable,
 				() -> mongoOperations.count(q, example.getProbeType(), entityInformation.getCollectionName()));
 	}
+	
+	/*
+	 * (non-Javadoc)
+	 * @see org.springframework.data.mongodb.repository.MongoRepository#findAllByExample(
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ketankkp27 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-mongodb&pullRequestId=518)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-mongodb&pullRequestId=518) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@ketankkp27 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-mongodb&pullRequestId=518)!

olivergierke:
Please open a ticket in our JIRA first to discuss your usecase. A `Query` instance provided *outside* of the repository doesn't make too much sense as it leaks store specifics into the client. A repository is supposed to abstract that away from clients.

PR #520, opened 2017-12-01
DATAMONGO-1831 - Fix array type conversion for empty source.

christophstrobl:
We now make sure that we convert empty sources to the correct corresponding target type. This prevents entity instantiation from failing due to incorrect argument types when invoking the constructor.

----


mp911de:

PR #521, opened 2017-12-11
DATAMONGO-1824 - Fix aggregation execution for MongoDB 3.6

christophstrobl:
Add `cursor` option to aggregation execution to make sure command works with MongoDB 3.6.

Tested against MongoDB Server `3.2.6`, `3.4.9` and `3.6.0`.

----

Should be forward ported to `2.0.x`. 
`2.1.x` (Lovelace) already contains a different kind of fix using the drivers `aggregate` method instead of executing the command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ jdk:
 before_script:
   - mongod --version
 
+services:
+  - mongodb
+
 env:
   matrix:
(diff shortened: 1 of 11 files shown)


mp911de:
That's now merged, polished and forward-ported to 2.0.x.

PR #526, opened 2018-01-16
DATAMONGO-1843 - Fix parameter shadowing in ArrayOperators reduce.

christophstrobl:
should also be back ported to `1.10.x` and `2.0.x`.


mp911de:
That's now merged, polished and backported.

PR #528, opened 2018-01-30
DATAMONGO-1803 - Add support for ChangeStreams

christophstrobl:
**Opened for Discussion**

**NOTE:** [Change Stream](https://docs.mongodb.com/manual/changeStreams/) support is only possible for replica sets or a sharded cluster.

Change Streams can be subscribed to with both the imperative and the reactive MongoDB java driver. It is highly recommended to use the reactive variant as it is less resource intensive. However if you do not feel comfortable using the reactive API for whatever reason, you can sill obtain the change events via a Messaging concept already common in the Spring ecosystem.

### Change Streams - Sync

Listening to a [Change Stream using a Sync Driver](https://docs.mongodb.com/manual/tutorial/change-streams-example/) is a long running, blocking task that needs to be delegated to a separate component.
In this case we need to create a `MessageListenerContainer` first which will be the main entry point for running the specific ``SubscriptionRequest``s.
Spring Data MongoDB already ships with a default implementation that operates upon `MongoTemplate` and is capable of creating and executing ``Task``s for a `ChangeStreamRequest`.

```java
MessageListenerContainer container = new DefaultMessageListenerContainer(template);
container.start();

MessageListener<Message<ChangeStreamDocument<Document>, User>> listener = System.out::println;
ChangeStreamRequestOptions options = new ChangeStreamRequestOptions("user", ChangeStreamOptions.empty());

Subscription subscription = container.register(new ChangeStreamRequest<>(listener, options), User.class);

// ...

container.stop();
```
### Change Streams - Reactive

Subscribing to Change Stream via the reactive API is clearly more straight forward. Still the building blocks like `ChangeStreamOptions` remain the same.

```java
Aggregation filter = newAggregation(User.class, match(where("age").gte(38));
Flux<ChangeStreamEvent<User>> flux = reactiveTemplate.changeStream(filter), User.class, ChangeStreamOptions.empty());
```


mp911de:
That's merged now.

PR #530, opened 2018-02-09
DATAMONGO-1865 - Avoid IncorrectResultSizeDataAccessException for derived findFirst/findTop queries and align behavior of reactive repository finder methods.

christophstrobl:
We now return the first result when executing `findFirst`/`findTop` queries. This fixes a glitch introduced in the Kay release throwing `IncorrectResultSizeDataAccessException` for single entity executions returning more than one result, which is explicitly not the desired behavior in this case.

Further more we aligned reactive repository finder invocations to follow the same `findOne` / `-Top` logic as the imperative counterpart.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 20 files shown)


mp911de:
That's merged, polished and back-ported now.

PR #532, opened 2018-02-13
2.0.x - DATAMONGO-1870 - Consider skip/limit on remove(Query, Class). 

christophstrobl:
We now use an `_id` lookup for remove operations using a `Query` with `limit` or `skip`. This allows more fine grained control over documents removed.

```java
// Remove a single entity via its id from the associated collection
template.remove(tywin, "GOT"); 

// Remove all documents matching the criteria of the query from the `GOT` collection.
template.remove(query(where("lastname").is("lannister")), "GOT");

// Remove the first 3 documents in the `GOT` collection. 
// Unlike <2> the documents to remove are identified via their `id` using the given query applying `sort`, `limit` and `skip` options and then removed all at once in a separate step.
template.remove(new Query().limit(3), "GOT");

// Remove all documents matching the criteria of the query from the `GOT` collection. 
// Unlike <3> documents do not get deleted in a batch but one by one.
template.findAllAndRemove(query(where("lastname").is("lannister"), "GOT");

// Remove the first 3 documents in the `GOT` collection. 
// Unlike <3> documents do not get deleted in a batch but one by one.
template.findAllAndRemove(new Query().limit(3), "GOT");
```

----

Should be forward ported to current _master_. 
For _1.10.x_ backport see #531

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>2.0.4.BUILD-SNAPSHOT</version>
+	<version>2.0.4.DATAMONGO-1870-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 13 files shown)


mp911de:
That's merged, polished, and ported to master now.

PR #533, opened 2018-02-13
DATAMONGO-1871 - Fix AggregationExpression aliasing. 

christophstrobl:
We now make sure to allow nested property alias by setting the target.


mp911de:

PR #538, opened 2018-03-20
DATAMONGO-1893 - Allow exclusion of fields other than _id in aggregation $project.

christophstrobl:
As of MongoDB 3.4 exclusion of fields other than `_id` is allowed, so we removed the limitation in our code.

----


mp911de:
That's polished, merged and backported now.

PR #539, opened 2018-03-20
DATAMONGO-1834 - Add Mongo 3.6 support for DateOperators including timezones

yinzara:
While the issue only requested support for $dateFromString, $dateFromParts and $dateToParts, I actually implemented the entirety of the Mongo 3.6 supported attributes related to dates.  I have added extensive test cases in DateOperatorsUnitTests that cover existing methods (as there were not existing test cases for DateOperators) and new methods.

This includes support for specifying "timezone" with nearly every date function available.  I created a new DateAggregationExpression abstract base class to handle this case and included warnings in the JavaDocs about pre3.6 incompatibilities.  

This required many additional methods being added to support an optional timezone and all aggregation expression classes to be modified.

Mongo does not by default support a "$quarter" operator on a date (to get the business quarter) however it can be easily achieved through conditionals so I added it as a "pseudo AggregationExpression (as it is still an aggregation expression).  It's extremely useful in aggregation pipelines that you want to do logic on the business quarter a specific date occurred in (see below).  If they ever decide to add it, we'll already have support in the library :)

I thought it was helpful to be able to reference the current date as part of the aggregation framework as it's extremely useful in the group pipeline to create conditionals.  This also helps to alleviate differences in how certain date elements are calculated between languages/libraries.  I created a new DateFactory interface (that allows for injecting of fixed dates in the case of tests) that defaults to calling "new java.util.Date()" when the query is executed.  This change also contemplates [this issue](https://jira.mongodb.org/browse/SERVER-23656) which calls for a "$$now" aggregation pipeline variable to be bound to the current date.  In the future our DateFactory could produce a binding to that variable.

For instance, if you wanted to sum the 'amount' field of records that's date field was in the current calendar week/quarter/month/year:
                 group("employee.id")
                .count().as("total_sold")            
                .sum(when(and(
                            valueOf(dateOf("item_date").isoWeek()).equalTo(currentDate().isoWeek()),
                            valueOf(dateOf("item_date").isoWeekYear()).equalTo(currentDate().isoWeekYear())
                    )).thenValueOf("amount").otherwise(0)).as("wtd")
                .sum(when(and(
                            valueOf(dateOf("item_date").month()).equalTo(currentDate().month()),
                            valueOf(dateOf("item_date").year()).equalTo(currentDate().year())
                    )).thenValueOf("amount").otherwise(0)).as("mtd")
                .sum(when(and(
                            valueOf(dateOf("item_date").quarter()).equalTo(currentDate().quarter()),
                            valueOf(dateOf("item_date").year()).equalTo(currentDate().year())
                    )).thenValueOf("amount").otherwise(0)).as("qtd")
                .sum(when(
                            valueOf(dateOf("item_date").year()).equalTo(currentDate().year())
                    ).thenValueOf("amount").otherwise(0)).as("ytd")

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AbstractAggregationExpression.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AbstractAggregationExpression.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016. the original author or authors.
+ * Copyright 2016-2018. the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.bson.Document;
 import org.springframework.util.ObjectUtils;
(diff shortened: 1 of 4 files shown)


christophstrobl:
thank you @yinzara! That's been merged, slightly polished and back ported.

PR #547, opened 2018-04-03
DATAMONGO-1916 - Fix potential ClassCastException in MappingMongoConverter.

christophstrobl:


mp911de:

PR #549, opened 2018-04-16
DATAMONGO-1889 - Add reactive group operation support.

christophstrobl:


mp911de:
We decided to drop this one as MongoDB deprecated `$group` and we don't want to introduce to-be-removed API on the reactive Template.

Closing without merging.

PR #551, opened 2018-04-23
DATAMONGO-1929 - Add fluent mapReduce template API.

christophstrobl:
Add ExecutableMapReduceOperation and ReactiveMapReduceOperation for fluent sync and reactive mapReduce support. Provide Kotlin extensions for both of them.


mp911de:
That's merged now.

PR #552, opened 2018-04-30
DATAMONGO-1963 - Allow DateOperatorFactory to be subclassed

yinzara:
This class prevents any type of subclassing because of the private constructor and fields which prevents people from creating their own operations on Mongo dates.  Requesting change of visibility to protected.

Adding additional pull request for 1.10.x branch as well.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/DateOperators.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/DateOperators.java
@@ -192,10 +192,10 @@ public class DateOperators {
 	 */
 	public static class DateOperatorFactory {
 
-		private final @Nullable String fieldReference;
-		private final @Nullable Object dateValue;
-		private final @Nullable AggregationExpression expression;
-		private final Timezone timezone;
+		protected final @Nullable String fieldReference;
+		protected final @Nullable Object dateValue;
+		protected final @Nullable AggregationExpression expression;
+		protected final Timezone timezone;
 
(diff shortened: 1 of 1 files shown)


yinzara:
https://github.com/spring-projects/spring-data-mongodb/pull/553

mp911de:
Please create a ticket in our [issue tracker](https://jira.spring.io/browse/DATAMONGO/) first so we can understand, what exactly you want to achieve.

In general, it's not necessary to submit pull requests for various branches. Typically, we take care of backports/forwardports during the merge.

yinzara:
Added issue https://jira.spring.io/browse/DATAMONGO-1963

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/DateOperators.java
```diff
@@ -192,10 +192,10 @@ public static Timezone ofExpression(AggregationExpression expression) {
 	 */
 	public static class DateOperatorFactory {
 
-		private final @Nullable String fieldReference;
-		private final @Nullable Object dateValue;
-		private final @Nullable AggregationExpression expression;
-		private final Timezone timezone;
+		protected final @Nullable String fieldReference;
```
mp911de: Let's keep these `private`, otherwise implementation details leak into subclasses. You can still keep these fields in your subclass.

mp911de:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #553, opened 2018-04-30
Allow DateOperatorFactory to be subclassed

yinzara:
This class prevents any type of subclassing because of the private constructor and fields which prevents people from creating their own operations on Mongo dates.  Requesting change of visibility to protected.

See other pull request to master:
https://github.com/spring-projects/spring-data-mongodb/pull/552/commits


yinzara:
Closed per request that only one branch is required in pull request.  See PR 552

PR #554, opened 2018-05-04
DATAMONGO-1920 - Add support for MongoDB 4.0 transactions (snyc) 

christophstrobl:
ℹ️ Requires: [MongoDB Server 3.7.9](https://www.mongodb.com/download-center#development) ℹ️ 

----

### MongoDB Transactions

As of version 4, MongoDB will support [Transactions](https://www.mongodb.com/transactions). Transactions are built on top of ``Session``s and therefore require an active `ClientSession`.

⚠️  Unless you specify a `MongoTransactionManager` within your application context, transaction support is **DISABLED**. You may use `setSessionSynchronization(ANY)` to participate in ongoing non-native MongoDB transactions.

To get full programmatic control over transactions, you may want to use the session callback on `MongoOperations`.

An example of programmatic transaction control within a `SessionCallback` is shown below:

**Programmatic transactions**

```java
ClientSession session = client.startSession(options); 

template.withSession(session)
    .execute(action -> {

        session.startTransaction();

        try {

            Step step = // ...;
            action.insert(step);

            process(step);

            action.update(Step.class).apply(Update.set("state", // ...

            session.commitTransaction();

        } catch (RuntimeException e) {
            session.abortTransaction();
        }
    }, ClientSession::close)
    .subscribe();
```

The above example allows you to have full control over transactional behavior while using the session scoped `MongoOperations` instance within the callback to ensure the session is passed on to every server call.
To avoid some of the overhead that comes with this approach usage of a `TransactionTemplate` can take away some of the noise of manual transaction flow.

**Transactions with TransactionTemplate**

```java
template.setSessionSynchronization(ANY);

// ...

TransactionTemplate txTemplate = new TransactionTemplate(anyTxManager);

txTemplate.execute(new TransactionCallbackWithoutResult() {

	@Override
	protected void doInTransactionWithoutResult(TransactionStatus status) {

		Step step = // ...;
		template.insert(step);

		process(step);

		template.update(Step.class).apply(Update.set("state", // ...
	};
});
```

**Transactions with MongoTransactionManager**

`MongoTransactionManager` is the gateway to the well known Spring transaction support. It allows applications to use [managed transaction features of Spring](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/transaction.html).
The `MongoTransactionManager` binds a `ClientSession` to the thread. `MongoTemplate` detects those and operates on these resources which are associated with the transaction accordingly. `MongoTemplate` can also participate in other, ongoing transactions.

```java
@Configuration
static class Config extends AbstractMongoConfiguration {

	@Bean
	MongoTransactionManager transactionManager(MongoDbFactory dbFactory) {
		return new MongoTransactionManager(dbFactory);
	}

	// ...
}

@Component
public class StateService {

	@Transactional
	void someBusinessFunction(Step step) {

		template.insert(step);

		process(step);

		template.update(Step.class).apply(Update.set("state", // ...
	};
});
```

⚠️ `@Transactional(readOnly = true)` advises `MongoTransactionManager` also to start a transaction adding the`ClientSession` to outgoing requests.


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoDatabaseUtils.java
```diff
@@ -0,0 +1,233 @@
...
+
+	@Nullable
+	private static ClientSession doGetSession(MongoDbFactory dbFactory, SessionSynchronization sessionSynchronization) {
+
+		MongoResourceHolder resourceHolder = (MongoResourceHolder) TransactionSynchronizationManager.getResource(dbFactory);
+
+		// check for native MongoDB transaction
+		if (resourceHolder != null && (resourceHolder.hasSession() || resourceHolder.isSynchronizedWithTransaction())) {
+
+			resourceHolder.requested();
```
mp911de: RefCount is incremented on each access. I think we should increment this one only on first access. Otherwise, multiple calls will reference the underlying resource multiple times without reducing refcount.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/SessionSynchronization.java
```diff
@@ -0,0 +1,37 @@
...
+package org.springframework.data.mongodb;
+
+/**
+ * {@link SessionSynchronization} is used along with {@link org.springframework.data.mongodb.core.MongoTemplate} to
+ * define in which type of transactions to participate if any.
+ *
+ * @author Christoph Strobl
+ * @since 2.1
+ */
+public enum SessionSynchronization {
```
mp911de: NATIVE and ANY are hard to guess names. How about `ALWAYS`, `ON_ACTUAL_TRANSACTION` and `NEVER` (to omit transactional participation even if a Tx is active)?

juergenzimmermann:
@christophstrobl Which artifact (and repo) is necessary to try you transactional implementation?

christophstrobl:
@juergenzimmermann 
```xml
<dependency>
    <groupId>org.springframework.data</groupId>
    <artifactId>spring-data-mongodb</artifactId>
    <version>2.1.0.DATAMONGO-1920-SNAPSHOT</version>
</dependency>
```

via 

```xml
<repositories>
    <repository>
        <id>spring-libs-snapshot</id>
        <url>https://repo.spring.io/libs-snapshot</url>
    </repository>
</repositories>
```

mp911de:
That's merged and polished now!

PR #555, opened 2018-05-07
DATAMONGO-1914 - Return an empty GridFsResource instead of null

christophstrobl:


mp911de:

PR #557, opened 2018-05-09
DATAMONGO-1968 - Add configuration support for com.mongodb.client.MongoClient

christophstrobl:
As of MongoDB java driver 3.7.0 there is an alternative entry point to `MongoClient` via the `mongodb-driver-sync` artifact.
`com.mongodb.client.MongoClient` is *not* compatible with `com.mongodb.MongoClient` and does not longer support the legacy `DBObject` codec. Therefore it cannot be used with `Querydsl` and requires a different configuration.
You may use `AbstractMongoClientConfiguration` to leverage the new `MongoClients` builder API.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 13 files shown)


mp911de:

PR #558, opened 2018-05-09
Fix minor typo in positionKey private field.

victor-bortone:
There is a minor typo in Field class (positionKey private field). This commit fixes it.


mp911de:
Thanks a lot. That's merged and backported now.

PR #559, opened 2018-05-10
FUll editing pass for Spring Data MongoDB

Buzzardo:
I did a full editing pass of the Spring Data MongoDB reference guide. I also adjusted index.adoc to work with the changes I made to the build project, so that we get Epub and PDF as well as HTML.


src/main/asciidoc/index.adoc
```diff
@@ -1,50 +1,48 @@
...
 (C) 2008-2018 The original authors.
 
-NOTE: _Copies of this document may be made for your own use and for distribution to others, provided that you do not charge any fee for such copies and further provided that each copy contains this Copyright Notice, whether distributed in print or electronically._
-
-toc::[]
+NOTE: Copies of this document may be made for your own use and for distribution to others, provided that you do not charge any fee for such copies and further provided that each copy contains this Copyright Notice, whether distributed in print or electronically.
 
 include::preface.adoc[]
 
-:leveloffset: +1
```
mp911de: What's the reason behind this change? We're using `:leveloffset:` in the other docs, too and didn't change these during editing?

mp911de:
Thanks a lot. That's merged, polished, and backported now.

src/main/asciidoc/index.adoc
Buzzardo: I discovered that epub generation through Asciidoctor has trouble with doing leveloffset that way but works if I put leveloffset=+1 in each include statement.

Asciidoctor's epub maker is still in alpha, so it has some limitations.

PR #560, opened 2018-05-11
DATAMONGO-1970 - Add support for MongoDB 4.0 transactions (reactive)

christophstrobl:
ℹ️ Requires: MongoDB Server 3.7.9  ℹ️ 

## Reactive transactions

Same as with the reactive `ClientSession` support, the `ReactiveMongoTemplate` offers dedicated methods for operating within a transaction without having to worry about commit/abort actions depending on the operations outcome.

`MongoOperations.inTransaction()` allows you to utilize the callback from for the reactive session support.

Everything happening inside the transactional callback is executed within a transaction. 
```java
template.inTransaction()
  .execute(action -> action.find(query(where("state").is("active")), Step.class)
      .flatMap(step ->
          action.update(Step.class)
              .matching(query(where("id").is(step.id))).apply(update("state", "paused")).all()
      )
  ).subscribe();
```
Errors subsequent in the reactive flow do not influence the transaction in any way like shown below

```java
template.inTransaction()
  .execute(action -> ...
  )
  .flatMap(result -> {
     // an error here as no affect on the transaction.
  })
  .subscribe();
```

**NOTE**: In case you need access to the `ClientSession` within the flow, you can use `ReactiveMongoContext.getSession()` to obtain in from the Reactor `Context`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,8 +28,8 @@
 		<project.type>multi</project.type>
 		<dist.id>spring-data-mongodb</dist.id>
 		<springdata.commons>2.1.0.BUILD-SNAPSHOT</springdata.commons>
-		<mongo>3.8.0-beta1</mongo>
-		<mongo.reactivestreams>1.8.0</mongo.reactivestreams>
+		<mongo>3.8.0-beta2</mongo>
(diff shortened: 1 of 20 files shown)


mp911de:

PR #561, opened 2018-05-14
DATAMONGO-1466 - Add embedded type information based reading GeoJSON converter

christophstrobl:
should be back ported to 1.x as well.


mp911de:
That's merged, polished and backported to 1.10.x and 2.0.x.

PR #564, opened 2018-05-24
DATAMONGO-1986 - Fix unmapped objects in aggregation pipeline.

christophstrobl:
We now initialize a `TypeBasedAggregationOperationContext` for ``TypedAggregation``s no matter what. This makes sure that potential `Criteria` objects are run trough the `QueryMapper`.
In case the default context is used we now also make sure to at least run the aggregation pipeline through the `QueryMapper` to avoid passing on non MongoDB simple types to the drive


mp911de:

PR #566, opened 2018-06-05
DATAMONGO-1979 - Add support for @Query(sort = "… 

christophstrobl:
We now allow to set a default sort for repository query methods via the `@Query` annotation.

```java
@Query(sort = "{ age : -1 }")
List<Person> findByFirstname(String firstname);
```

Using an explicit `Sort` parameter along with the annotated one allows to alter the defaults set via the annotation. Method argument sort parameters add to / override the annotated defaults.

```java
@Query(sort = "{ age : -1 }")
List<Person> findByFirstname(String firstname, Sort sort);
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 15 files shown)


mp911de:
That's polished and merged now.

PR #567, opened 2018-06-06
DATAMONGO-1998 - Fix Querydsl id handling for property references using ObjectId hex String representation.

christophstrobl:
We now follow the conversion rules for `id` properties with a valid `ObjectId` representation when parsing Querydsl queries.


mp911de:

PR #568, opened 2018-06-07
DATAMONGO-2001 - Count within session should return only the total count of documents visible to the specific session

christophstrobl:
MongoDB `count` operates upon collection statistics which may not reflect the actual situation within a transaction. Therefore the server (as of _4.0.0-rc3_) responds with _error 50851_ when issuing a `count` command inside of a multi-document transaction.
Once `MongoTemplate` detects an active transaction, all exposed `count(...)` methods delegate to the aggregation framework using `$match` and `$count` operators, preserving `Query` settings, such as `collation`.

The following snippet of `count` inside the session bound closure

```java
session.startTransaction();

template.withSession(session)
    .execute(action -> {
        action.count(query(where("state").is("active")), Step.class)
        ...
```
runs:
```javascript
db.collection.aggregate(
   [
      { $match: { state: "active" } },
      { $group: { _id: null, count: { $sum: 1 } } }
   ]
)
```
instead of:
```javascript
db.collection.find( { state: "active" } ).count()
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,12 @@ before_install:
   - |-
     downloads/mongodb-linux-x86_64-ubuntu1604-${MONGO_VERSION}/bin/mongo --eval "rs.initiate({_id: 'rs0', members:[{_id: 0, host: '127.0.0.1:27017'}]});"
     sleep 15
-    
+
 env:
   matrix:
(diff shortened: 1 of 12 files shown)


mp911de:

PR #570, opened 2018-06-11
DATAMONGO-2003 - Fix derived query using regex pattern with options.

christophstrobl:
We now consider regex `Pattern` options when using the `Pattern` as a derived finder argument.


mp911de:

PR #572, opened 2018-06-14
DATAMONGO-1919 - Allow reactive-only usage of ReactiveMongoTemplate.

mp911de:
We now support reactive-only usage of Spring Data MongoDB without the need to use the synchronous driver or even having it on the class path. We conditionally register `CodeWScope`/`CodeWithScope` types that are bundled with the particular driver distributions.

`NoOpDbRefResolver` is now a top-level type to be used with a reactive-only `MongoMappingContext`.

---

Related ticket: [DATAMONGO-1919](https://jira.spring.io/browse/DATAMONGO-1919).


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/MongoSimpleTypes.java
```diff
@@ -37,31 +36,38 @@
...
 		simpleTypes.add(DBRef.class);
 		simpleTypes.add(ObjectId.class);
 		simpleTypes.add(BsonObjectId.class);
-		simpleTypes.add(CodeWScope.class);
 		simpleTypes.add(org.bson.Document.class);
 		simpleTypes.add(Pattern.class);
 		simpleTypes.add(Binary.class);
 		simpleTypes.add(UUID.class);
 
+		if (ClassUtils.isPresent("org.bson.types.CodeWScope", MongoSimpleTypes.class.getClassLoader())) {
```
christophstrobl: Looking at the latest driver bits, those changes here do not seem to be necessary any longer. Comparing the `org.bson` packages of `org.mongodb:mongo-java-driver:3.8.0` and `org.mongodb:bson:3.8.0` shows no difference at all.

I think we should wait until #576 is resolved.

PR #574, opened 2018-06-19
DATAMONGO-2005 - Switch reactive transactions API over to Flux.usingWhen(...

christophstrobl:
Requires Reactor `3.2.0.M2`.


mp911de:

PR #582, opened 2018-07-06
DATAMONGO-2022 - Refactor getResources(String locationPattern) for a …

nhh:
…more readable approach

The solution should be more readable, avoid using multiple additional dependencies
and decrease variable instantiation.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
@@ -15,14 +15,12 @@
  */
 package org.springframework.data.mongodb.gridfs;
 
-import static org.springframework.data.mongodb.core.query.Query.*;
-import static org.springframework.data.mongodb.gridfs.GridFsCriteria.*;
-
-import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
-
+import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.gridfs.GridFSBucket;
+import com.mongodb.client.gridfs.GridFSBuckets;
+import com.mongodb.client.gridfs.GridFSFindIterable;
+import com.mongodb.client.gridfs.model.GridFSFile;
+import com.mongodb.client.gridfs.model.GridFSUploadOptions;
 import org.bson.BsonObjectId;
 import org.bson.Document;
 import org.bson.types.ObjectId;
@@ -35,12 +33,12 @@ import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
-import com.mongodb.client.MongoDatabase;
-import com.mongodb.client.gridfs.GridFSBucket;
-import com.mongodb.client.gridfs.GridFSBuckets;
-import com.mongodb.client.gridfs.GridFSFindIterable;
-import com.mongodb.client.gridfs.model.GridFSFile;
-import com.mongodb.client.gridfs.model.GridFSUploadOptions;
+import java.io.InputStream;
+import java.util.Optional;
+import java.util.stream.StreamSupport;
+
+import static org.springframework.data.mongodb.core.query.Query.query;
+import static org.springframework.data.mongodb.gridfs.GridFsCriteria.whereFilename;
 
 /**
  * {@link GridFsOperations} implementation to store content into MongoDB GridFS.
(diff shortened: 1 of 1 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
```diff
@@ -247,28 +246,15 @@ public GridFsResource getResource(GridFSFile file) {
...
-		if (!StringUtils.hasText(locationPattern)) {
-			return new GridFsResource[0];
-		}
+    public GridFsResource[] getResources(String locationPattern) {
+        Pattern pattern = Pattern.compile(locationPattern);
 
-		AntPath path = new AntPath(locationPattern);
+        return StreamSupport
+                .stream(find(query(whereFilename().regex(pattern))).spliterator(), false)
+                .map(GridFsResource::new)
```
olivergierke: This invocation seems to result in something different than the old constructor invocation. The `getGridFs().…` call is missing. Is this an issue?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
```diff
@@ -247,28 +246,15 @@ public GridFsResource getResource(GridFSFile file) {
      * (non-Javadoc)
      * @see org.springframework.core.io.support.ResourcePatternResolver#getResources(java.lang.String)
      */
-	public GridFsResource[] getResources(String locationPattern) {
-
-		if (!StringUtils.hasText(locationPattern)) {
-			return new GridFsResource[0];
-		}
+    public GridFsResource[] getResources(String locationPattern) {
+        Pattern pattern = Pattern.compile(locationPattern);
```
olivergierke: The new code skips treating the location as `AntPath` and expects it to be a regex expression right away. This is clearly breaking the contract, isn't it?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
nhanft: Yes, you are right, i misused the possibility to instatiate a GridFsResource with a GridFsFile. I thought this would also provide a valid GridFsResource, but as i looked into it seems that the inputStream is missing. I will have a look. Thx for your feedback!

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
nhh: Yes, you are right, i misused the possibility to instatiate a GridFsResource with a GridFsFile. I thought this would also provide a valid GridFsResource, but as i looked into it seems that the inputStream is missing. I will have a look. Thx for your feedback!

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
nhh: Yes, you are right, i misused the possibility to instatiate a GridFsResource with a GridFsFile. I thought this would also provide a valid GridFsResource, but as i looked into it seems that the inputStream is missing. I will have a look. Thx for your feedback!

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
nhh: Yes, it is. I will come up with a new solution, working on with your feedback.

nhh:
@olivergierke I changed some of the code, is this more suitable?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/GridFsTemplate.java
```diff
@@ -15,14 +15,12 @@
  */
 package org.springframework.data.mongodb.gridfs;
 
-import static org.springframework.data.mongodb.core.query.Query.*;
```
mp911de: Please do not reformat the imports. Please find the import settings [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide).

mp911de:
Closing as we want to stick with the current approach. Using Java 8 stream imposes another layer of indirection and performance penalty.

PR #584, opened 2018-07-08
DATAMONGO-2024 adding jmx connection pool listener toggle to client options schema

steve-thousand:
I'm sure there may be better approaches to this than the one I chose, which was just a separate boolean for the one connection pool listener type. Maybe there is a more flexible option?

Also, far be it from me to propose a whole new schema version, but I wasn't sure if it was best practice to add the new option to the existing schema, or create a new version.

This is only my second contribution, and my first was mostly a one-line change, so I am very open to the possibility that this may have gone a little farther than it should have.


steve-thousand:
This may not be entirely necessary at least for my purposes anymore. On the linked JIRA ticket, an alternate solution was suggested. I will give that a shot and see if it fixes my problem

steve-thousand:
Verified that the solution proposed on the ticket worked. We simply added a BeanPostProcessor to catch the mongoClientOptions and add a jmx listener

PR #589, opened 2018-08-06
DATAMONGO-2043 - Omit type hint when mapping simple types.

christophstrobl:
Make sure to omit the type hint (aka `_class` attribute) when writing objects that are actually considered simple types, eg. `org.bson.Document`.


mp911de:

PR #590, opened 2018-08-07
DATAMONGO-2044 make ensureNotIterable actually check if object is iterable

GotoFinal:
Other option would be just to remove that check, as it might be possible that someone is saving objects that implement iterable but are valid data classes?

As current code just does nothing, as object can be of type that is implementing one of that interfaces, but it can not be of interface type itself.


GotoFinal:
Now I'm confused as now I see there is a test that ensures that iterable item will pass here, so maybe this code should be just removed?

christophstrobl:
thanks @GotoFinal. Sorry it took so long - merged to _main_ line.

PR #591, opened 2018-08-08
DATAMONGO-2041 - Apply field restriction to DTO projections.

christophstrobl:


mp911de:

PR #592, opened 2018-08-09
DATAMONGO-2045 - Add session & transaction specific error codes for exception translation.

christophstrobl:


mp911de:

PR #593, opened 2018-08-09
DATAMONGO-2047 - Update $dateToString and $dateFromString aggregation operators to match  MongoDB 4.0 changes.

christophstrobl:
We added the format option to `DateFromString` and made it optional for `DateToString`.

----

May be a candidate to port back to 2.0.x and 1.10.x.


mp911de:

PR #598, opened 2018-08-13
DATAMONGO-2051 - Add support for SCRAM-SHA-256 authentication mechanism to MongoCredentialPropertyEditor.

christophstrobl:


mp911de:

PR #604, opened 2018-08-17
DATAMONGO-2059 - Favor collection.countDocuments() over collection.count() which will be removed.

christophstrobl:
This commit switches from simple `collection.count()`, operating on potentially false collection statistic,  to `countDocuments()` using an aggregation for accurate results.
The transition required query modifications at some points because `$match` does not support `$near` nor `$nearSphere` but requires `$geoWithin` using `$center` or `$centerSphere` which does not support `$minDistance` (see [SERVER-37043](https://jira.mongodb.org/browse/SERVER-37043)).
`$geoWithin` further more does not sort results by distance, but this fact can be ignored when just counting matches.

ℹ️ The change is a requirement for upgrading to the next driver version that no longer supports `collection.count()`.

**Examples:**

```json
// before
{ "location" : { "$near" : [-73.99171, 40.738868], "$maxDistance" : 1.1 } }

// after
{ "location" : { "$geoWithin" : { "$center" : [ [-73.99171, 40.738868], 1.1] } } }
```
```json
// before
{ "location" : { "$near" : [-73.99171, 40.738868], "$minDistance" : 0.1, "$maxDistance" : 1.1 } }

// after
{ "$and" :[ 
    { "$nor" :[ { "location" :{ "$geoWithin" :{ "$center" :[ [-73.99171, 40.738868 ], 0.01] } } } ] },
    { "location" :{ "$geoWithin" :{ "$center" :[ [-73.99171, 40.738868 ], 1.1] } } } 
] }
```


doronlevari:
Can we get this one merged into 2.2.X? It doesn't seem THAT major and we're in a huge need for this, Mongo is totally crapping out on us when using "count"

PR #605, opened 2018-08-28
Refine exception translation for client session exceptions

christophstrobl:
We now distinguish between `Transient` and `NonTransient` failures by checking the Error labels of an Error and create the according `DataAccessException` based on that information.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
@@ -25,9 +25,11 @@ import org.springframework.dao.DuplicateKeyException;
 import org.springframework.dao.InvalidDataAccessApiUsageException;
 import org.springframework.dao.InvalidDataAccessResourceUsageException;
 import org.springframework.dao.PermissionDeniedDataAccessException;
+import org.springframework.dao.TransientDataAccessException;
 import org.springframework.dao.support.PersistenceExceptionTranslator;
 import org.springframework.data.mongodb.ClientSessionException;
-import org.springframework.data.mongodb.MongoTransactionException;
+import org.springframework.data.mongodb.TransientClientSessionException;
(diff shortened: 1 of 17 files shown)


mp911de:
Spring's DAO exceptions do not differentiate between contextual transience in the sense that MongoDB does. Re-wrapping exceptions to add another transient flavour isn't appropriate. We also do not have a way to add this flavor to exceptions that we use such as `DataAccessResourceFailureException`. It's always possible to introspect the original MongoDB exception labels, especially that we now provide `MongoExceptionTranslator.isTransientFailure`.


I updated the PR to allow the configuration of a `MongoExceptionTranslator` so that applications can tailor exception translation to their requirements.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
```diff
@@ -115,20 +124,34 @@ public DataAccessException translateExceptionIfPossible(RuntimeException ex) {
...
+				return new DataIntegrityViolationException(mongoException.getMessage(), mongoException);
+			}
+			if (MongoDbErrorCodes.isClientSessionFailure(mongoException)) {
+				return isTransientFailure(mongoException) ? new TransientClientSessionException(ex.getMessage(), ex)
+						: new ClientSessionException(ex.getMessage(), ex);
+			}
+			if (MongoDbErrorCodes.isDataIntegrityViolationError(mongoException)) {
+				return new DataIntegrityViolationException(mongoException.getMessage(), mongoException);
+			}
+			if (MongoDbErrorCodes.isClientSessionFailure(mongoException)) {
```
christophstrobl, line 150: same as in line 143?

mp911de: the `DataIntegrityViolation` block is duplicated as well, seems a remainder from conflict resolution. Both `if` branches can go away.

PR #606, opened 2018-08-29
DATAMONGO-2075 - Open up MongoTransaction manager to allow transaction commit retry.

christophstrobl:
MongoDB can add special labels to errors raised during a transactional execution. Those may indicate transient failures that might vanish by simply retrying the operation.
We highly recommend [Spring Retry](https://github.com/spring-projects/spring-retry) for those purposes. Nevertheless one may override `MongoTransactionManager#doCommit(MongoTransactionObject)` to implement a [Retry Commit Operation](https://docs.mongodb.com/manual/core/transactions/#retry-commit-operation)
behavior as outlined in the MongoDB reference manual.

```java
@Bean
MongoTransactionManager txManager(MongoDbFactory dbFactory) {
  return new MongoTransactionManager(dbFactory) {
    @Override
    protected void doCommit(MongoTransactionObject transactionObject) throws Exception {
      int retries = 3;
      do {
        try {
          transactionObject.commitTransaction();
          break;
        } catch (MongoException ex) {
          if (!ex.hasErrorLabel(MongoException.UNKNOWN_TRANSACTION_COMMIT_RESULT_LABEL)) {
            throw ex;
          }
        }
        Thread.sleep(500);
      } while (--retries > 0);
    }
  };
}
```

### DIFF
--- a/src/main/asciidoc/reference/client-session-transactions.adoc
+++ b/src/main/asciidoc/reference/client-session-transactions.adoc
@@ -315,6 +315,14 @@ MongoDB does *not* support collection operations, such as collection creation, w
 affects the on the fly collection creation that happens on first usage. Therefore make sure to have all required
 structures in place.
 
+*Transient Errors*
+
+MongoDB can add special labels to errors raised during a transactional execution. Those may indicate transient failures
+that might vanish by simply retrying the operation.
+We highly recommend https://github.com/spring-projects/spring-retry[Spring Retry] for those purposes. Nevertheless
+one may override `MongoTransactionManager#doCommit(MongoTransactionObject)` to implement a https://docs.mongodb.com/manual/core/transactions/#retry-commit-operation[Retry Commit Operation]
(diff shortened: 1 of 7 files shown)


mp911de:

src/main/asciidoc/reference/client-session-transactions.adoc
```diff
@@ -315,6 +315,14 @@ MongoDB does *not* support collection operations, such as collection creation, w
 affects the on the fly collection creation that happens on first usage. Therefore make sure to have all required
 structures in place.
 
+*Transient Errors*
+
+MongoDB can add special labels to errors raised during a transactional execution. Those may indicate transient failures
+that might vanish by simply retrying the operation.
+We highly recommend https://github.com/spring-projects/spring-retry[Spring Retry] for those purposes. Nevertheless
```
LEQADA, line 322: On which exact exceptions should one retry? Retrying on `MongoException` like in the example from the PR description will probably backfire as it's too broad.

PR #613, opened 2018-10-01
DATAMONGO-2096 - Fix target field name for GraphLookup aggregation operation.

christophstrobl:
We mow make sure to use the target field name instead of the alias when processing `GraphLookupOperation`.


mp911de:
That's polished, merged, and backported now.

PR #614, opened 2018-10-02
DATAMONGO-2101 - Fix DBObject to GeoJson conversion. 

christophstrobl:
Querydsl still wraps MongoDB data in `DBObject` which causes trouble with the registered converters that deal with from `Document` conversion. Therefore we now try to extract the argument map from the `DBObject` transferring it to `Document` in order to have the converters kick in where applicable.

This issue applies to the 2.0.x line only as we added Querydsl support for `Document` in `2.1` and use `DBObject` in `1.x`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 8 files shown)


mp911de:
That's merged to `2.0.x` now.

PR #615, opened 2018-10-25
DATAMONGO-2113 - Fix resumeTimestamp conversion for change streams.

christophstrobl:
We now use the first 32 bits of the `BsonTimestamp` to create the `Instant` and ignore the ordinal value.


mp911de:

PR #617, opened 2018-11-06
DATAMONGO-1798 - Introduce @MongoId annotation for id conversion control.

christophstrobl:
`@MongoId` allows more fine grained control over id conversion by specifying the intended id target type. This allows to skip the automatic to ObjectId conversion of values that happen to be valid ObjectId hex strings.

```java
class PlainStringId {
  @MongoId String id; // treated as String no matter what
}

class PlainObjectId {
  @MongoId ObjectId id; // treated as ObjectId
}

class StringToObjectId {
  @MongoId(ObjectId.class) String id; // treated as ObjectId if the value is a valid ObjectId hex string
}
```


mp911de:

PR #618, opened 2018-11-13
DATAMONGO-2130 - Fix Repository count inside transaction.

christophstrobl:
We now make sure invocations on repository `count` methods delegate to `countDocuments` when inside a transaction.


mp911de:

PR #623, opened 2018-11-29
DATAMONGO-2149 - Fix $slice in fields projection when pointing to array of DBRefs.

christophstrobl:
We now no longer try to convert the actual slice parameters into a `DBRef`.

--- 

Should be back ported to _2.0.x_ and _2.1.x_. Maybe even _1.x_


mp911de:

PR #624, opened 2018-11-30
DATAMONGO-2115 - Add support for BsonTimestamp when resuming Change Streams.

christophstrobl:


mp911de:

PR #625, opened 2018-12-04
DATAMONGO-2155 - Fix mapping of Map key/value pairs.

christophstrobl:
We now make sure to map key/value pairs of `Map` like properties to the values domain type, and apply potentially registered custom converters to the keys.

We also fixed an invalid test for [DATAMONGO-1423](https://jira.spring.io/browse/DATAMONGO-1423) as this one did not check the application of the registered converter and now bypass a 2nd round of (actually not needed) conversion when the update has been generated by the template API to persist a versioned entity.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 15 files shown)


mp911de:
That's polished, merged, and backported now.

PR #630, opened 2018-12-27
Fixed broken auditing for entities on batch inserts.

zhujk:
Fixed broken auditing for entities on batch inserts.


mp911de:
Thanks for submitting a pull request. We will fix the bug along with the ticket [DATAMONGO-2180](https://jira.spring.io/browse/DATAMONGO-2180) that also fixes the same issue in the imperative implementation. Therefore, we're not going to consider this pull request and we're closing this one without merging.

PR #633, opened 2019-01-09
 DATAMONGO-2174 - Fix InvalidPersistentPropertyPath exception when updating documents.

christophstrobl:
`MetadataBackedField` now returns `null` instead throwing an error for fields that are not part of the domain model. This allows adding any field when updating an entity.


mp911de:

PR #634, opened 2019-01-09
DATAMONGO-2173 - Translate and forward Exception during CursorReadingTask#start() to ErrorHandler.

christophstrobl:
We now make sure to translate and pass on errors during the cursor initialization procedure to the configured `ErrorHandler`.


mp911de:

PR #640, opened 2019-01-30
DATAMONGO-2193 - Fix String <> ObjectId conversion for non id properties.

christophstrobl:
We now make sure to only convert valid `ObjectId` Strings if the property can be considered as _id _property.

----

Should be forward ported to master.


mp911de:
That's merged, polished and forward-ported.

PR #641, opened 2019-02-04
DATAMONGO-2195 - Consider version when removing an entity.

christophstrobl:
We now consider a potential `@Version` when removing an entity.

`MongoOperations#remove(Object)` and `MongoOperations#remove(Object,  String)` now throw an `OptimisticLockingFailureException` when the object to remove is versioned and the version of the given entity does not match the one within the store. If no document with matching `_id` can be found the remove operation passes without an error. Same applies to _remove_ operations executed inside an ongoing MongoDB transaction.

`MongoRepository` now also considers the entities version, following the very same logic as `MongoOperations`.

To remove an entity without version check use `MongoOperations#remove(Query,…)` or `MongoRepository#deleteById(…`

This PR also includes a fix (2nd commit) for [DATAMONGO-2196](https://jira.spring.io/browse/DATAMONGO-2196) that should be back ported to _2.1.x_ and _2.0.x_.


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ReactiveMongoTemplate.java
```diff
@@ -1822,10 +1821,47 @@ private void assertUpdateableIdIfNotSet(Object value) {
 							return collectionToUse.deleteMany(MappedDocument.getIdIn(val), deleteOptions);
 						});
 			} else {
-				return collectionToUse.deleteMany(removeQuey, deleteOptions);
+				return collectionToUse.deleteMany(removeQuery, deleteOptions);
+			}
+
+		}).flatMap(deleteResult -> {
+
+			if (entity != null && entity.hasVersionProperty()) {
```
odrotbohm: Please invert the if clauses to use early returns to avoid the heavy nesting.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/EntityOperations.java
```diff
@@ -199,6 +199,21 @@ private static Document parse(String source) {
...
 
+		/**
+		 * Returns the {@link Query} to remove an entity by its {@literal id} and if applicable {@literal version}.
+		 *
+		 * @return the {@link Query} to use for removing the entity. Never {@literal null}.
+		 * @since 2.2
+		 */
+		default Query getRemoveByQuery() {
+
+			if (isVersionedEntity()) {
```
odrotbohm: Ternary ? true : false. ;)

mp911de:
That's merged, polished and partially backported now.

PR #643, opened 2019-02-14
DATAMONGO-2198 - Upgrade to MongoDB Java Driver 3.10.

christophstrobl:
With the upgrade to MongoDB Java Driver 3.10.x we already prepare for the upcoming 4.0. release of the driver. Therefore we started moving away from deprecated API and introduced an alternative for the JSON parser used for constructing queries from plain String expressions used via the `@Query` annotation.


mp911de:

PR #644, opened 2019-02-18
DATAMONGO-1854 - Add support for Collation to repositories.

christophstrobl:
We now allow to specify the collation via the `@Query` annotation or a `Collation` method argument.

```java
public interface PersonRepository extends MongoRepository<Person, String> {

    @Query(collation = "en_US")
    List<Person> findByFirstname(String firstname);

    @Query(collation = "{ 'locale' : 'en_US' }")
    List<Person> findPersonByFirstname(String firstname);

    @Query(collation = "?1")
    List<Person> findByFirstname(String firstname, Object collation);

    @Query(collation = "{ 'locale' : '?1' }")
    List<Person> findByFirstname(String firstname, String collation);

    List<Person> findByFirstname(String firstname, Collation collation);
}
```

> ℹ️ In case auto index creation is enabled for repository query methods, fixed collation values like the ones of the first 2 methods of the example above, are added to the index options.


mp911de:

PR #645, opened 2019-02-19
DATAMONGO-2072 - Support Range in repository between queries.

christophstrobl:
We now support `o.s.d.domain.Range` as method parameter for _between_ queries. This allows more fine grained control over the inclusion/exclusion of the upper/lower bounds.
Up till now _between_ required 2 parameters which had been strictly bound to excluding bounds using `$gt` and `$lt`.


mp911de:

PR #650, opened 2019-02-26
DATAMONGO-1783 - Apply query limit and offset to MongoOperations#count

christophstrobl:
We now pass on the range defined by `Query.skip` and `Query.limit` to the MongoDB server. This allows to count documents within a certain range so that it is possible to find the number of matches within eg. the first 10,000 documents.


mp911de:

PR #652, opened 2019-03-01
DATAMONGO-1593 - Fix JSON parsing for $oid in String based queries.

christophstrobl:
We now parse `String` arguments bound to an `$oid` in a way that creates a valid `ObjectId`.

```java
@Query("{ 'arg0' : { '$oid' : ?0 } }")
List<Person> useObjectIdInQueryString(String arg0);
```


mp911de:
That's merged now.

PR #653, opened 2019-03-01
DATAMONGO-2054 - Add support for array update operator $[].

christophstrobl:
We now support the `$[]` array update operator when mapping `Update`. The mapper now takes potential `@Field` annotations into account and applies the field name correctly.


mp911de:

PR #659, opened 2019-03-11
DATAMONGO-2224 - Add trace logging for DBRef resolution. 

christophstrobl:
We added trace logging to `DefaultDbRefResolver`.

```xml
<logger name="org.springframework.data.mongodb.core.convert.DefaultDbRefResolver" level="trace"/>
```

PR #660, opened 2019-03-11
DATAMONGO-2223 - Add test for DBRef resolution in a different database.

christophstrobl:
Includes #659. Please drop 2nd commit (bafbf51) when merging.


mp911de:
That's merged now.

PR #732, opened 2019-03-26
DATAMONGO-2221 - Fix mapping of Strings matching a valid ObjectId for unresolvable paths.

christophstrobl:
We now make sure we do **not** convert ``String``s that represent valid ``ObjectId``s into the such for paths that cannot be resolved to a `Property`.


mp911de:

PR #733, opened 2019-03-29
DATAMONGO-1849 - Resolve JsonSchema from domain type.

christophstrobl:
`JsonSchemaCreator` extracts the `MongoJsonSchema` for a given `Class` by applying the following mapping rules:

**Required Properties**
- All Constructor arguments unless annotated with Nullable.
- Properties of primitive type.

**Ignored Properties**
- All properties annotated with Transient.

**Property Type Mapping:**
- java.lang.Object -> `{ type : 'object' }`
- java.util.Arrays -> `{ type : 'array' }`
- java.util.Collection -> `{ type : 'array'}`
- java.util.Map -> `{ type : 'object'}`
- java.lang.Enum -> `{ type : 'string', enum : [ ... ] }`
- Simple Types -> `{ type : 'the corresponding bson type' }`
- Domain Types -> `{ type : 'object', properties : { ... } }`

`_id` properties using types that can be converted into `ObjectId` like `String` will be mapped to `{ type : 'object' }` unless there is more specific information available via the `MongoId` annotation.


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MappingMongoJsonSchemaCreator.java
```diff
@@ -0,0 +1,213 @@
...
+
+	private String computePropertyFieldName(PersistentProperty property) {
+
+		return property instanceof MongoPersistentProperty ? ((MongoPersistentProperty) property).getFieldName()
+				: property.getName();
+	}
+
+	private boolean isRequiredProperty(PersistentEntity<?, ?> parent, PersistentProperty property) {
+
+		return (parent.isConstructorArgument(property) && !property.isAnnotationPresent(Nullable.class))
```
mp911de: See `NullableUtils`: Properties are nullable by default. Annotating a package with  `@NonNullApi` or `@NonNullFields` switches to non-null by default.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoJsonSchemaCreator.java
```diff
@@ -0,0 +1,73 @@
...
+	 */
+	MongoJsonSchema createSchemaFor(Class<?> type);
+
+	/**
+	 * Creates a new {@link MongoJsonSchemaCreator} that is aware of conversions applied by the given {@link MongoConverter}.
+	 *
+	 * @param mongoConverter must not be {@literal null}.
+	 * @return new instance of {@link MongoJsonSchemaCreator}.
+	 */
+	default MongoJsonSchemaCreator jsonSchemaCreator(MongoConverter mongoConverter) {
```
mp911de: Shouldn't this be `static`?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
```diff
@@ -1667,4 +1689,91 @@ private static boolean isCollectionOfDbRefWhereBulkFetchIsPossible(Iterable<?> s
 			return null;
 		}
 	}
+
+	private static class HijackedTypeInformation<S> implements TypeInformation<S> {
```
mp911de: How about `TypeInformationWrapper` 😬and handling the rest of customizations in a separated, smaller type?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MongoConverter.java
```diff
@@ -149,4 +149,15 @@ default Object convertId(@Nullable Object id, Class<?> targetType) {
...
 	}
+
+	/**
+	 * Compute the target type for a given source considering {@link org.springframework.data.convert.CustomConversions}.
+	 *
+	 * @param source the source type.
+	 * @return never {@literal null}.
+	 * @since 2.2
+	 */
+	default Class<?> computeWriteTarget(Class<?> source) {
```
mp911de: We should improve either on the definition or on the implementation of this method. Right now, we could assume that the write target is either a native Mongo type or the write target is an to be converted type.

The number of exposed methods via `MongoWriter` and `MongoConverter` keeps constantly increasing and I'm not sure how to think about it.

mp911de:

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MongoConverters.java
```diff
@@ -75,7 +76,9 @@ static Collection<Object> getConvertersToRegister() {
 		List<Object> converters = new ArrayList<>();
 
 		converters.add(BigDecimalToStringConverter.INSTANCE);
+		converters.add(BigDecimalToDecimal128Converter.INSTANCE);
```
mariusleu, line 79: This has no effect, as `BigDecimalToStringConverter.INSTANCE` from above always takes precedence. All BigDecimal end up as String in the database.

Why do we still need `BigDecimalToStringConverter`?

PR #735, opened 2019-04-03
DATAMONGO-2247 - Add non-nullable variants to Coroutines extensions

sdeleuze:


mp911de:
That's merged and polished now. Thanks a lot.

PR #744, opened 2019-05-02
DATAMONGO-2264 - Support upcoming MongoDB 4.2 Server.

christophstrobl:
Deprecate and guard tests for removed commands/options:
- eval
- group
- maxScan

Transition from removed `geoNear` command to `$geoNear` aggregation pipeline stage.
This allows users to still use `NearQuery` along with the `MongoOperations.near(...)` that now transforms the given query into a `GeoNearOperation` applying `$skip` (previously in memory skipping) and `$limit` (previous num parameter) as additional aggregation pipeline stages.

---

Tested against:
- 4.1.10
- 4.0.4
- 3.6.12
- 4.4.20

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,12 @@ before_install:
 
 env:
   matrix:
-    - PROFILE=ci
+    - MONGO_VERSION=4.1.10
+    - MONGO_VERSION=4.0.4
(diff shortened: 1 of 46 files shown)


mp911de:

PR #746, opened 2019-05-13
DATAMONGO-2267 - Fix eager collection resolution in Object path. 

christophstrobl:
We now lazily read the collection of an entity as it potentially requires a more expensive SpEL evaluation that might not have been required in fist place.


mp911de:

PR #743, opened 2019-04-18
DATAMONGO-2153 - Aggregation support for repositories.

christophstrobl:
The repository layer offers means interact with the aggregation framework via annotated repository finder methods. Similar to the JSON based queries a pipeline can be defined via the `@Aggregation` annotation. The definition may contain simple placeholders like `?0` as well as SpEL expression markers `?#{ ... }`.

**Sync**
```java
public interface PersonRepository extends CrudRepository<Person, String> {

  @Aggregation("{ $group: { _id : $lastname, names : { $addToSet : $?0 } } }")
  List<PersonAggregate> groupByLastnameAnd(String property);

  @Aggregation("{ $group: { _id : $lastname, names : { $addToSet : $firstname } } }")
  List<PersonAggregate> groupByLastnameAndFirstnames(Sort sort);

  @Aggregation("{ $group: { _id : $lastname, names : { $addToSet : $?0 } } }")
  List<PersonAggregate> groupByLastnameAnd(String property, Pageable page);

  @Aggregation("{ $group : { _id : null, total : { $sum : $age } } }")
  SumValue sumAgeUsingValueWrapper();

  @Aggregation("{ $group : { _id : null, total : { $sum : $age } } }")
  Long sumAge();

  @Aggregation("{ $group : { _id : null, total : { $sum : $age } } }")
  AggregationResults<SumValue> sumAgeRaw();

  @Aggregation("{ '$project': { '_id' : '$lastname' } }")
  List<String> findAllLastnames();
}
```
**Reactive**
```java
public interface ReactivePersonRepository extends ReactiveCrudReppsitory<Person, String> {

  @Aggregation("{ $group: { _id : $lastname, names : { $addToSet : $?0 } } }")
  Flux<PersonAggregate> groupByLastnameAnd(String property);

  @Aggregation("{ $group : { _id : null, total : { $sum : $age } } }")
  Mono<Long> sumAge();

  @Aggregation("{ '$project': { '_id' : '$lastname' } }")
  Flux<String> findAllLastnames();
}
```

PR #748, opened 2019-05-15
DATAMONGO-2200 - Derive fields for aggregation $project stage from a given type.

christophstrobl:
We now allow to derive field names for a `$project` stage from a given type by including all top level fields.

    // $project : { title : 1, author : 1 }
    Aggregation.project(Book.class)

Requires changes from: #743

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/ProjectionOperation.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/ProjectionOperation.java
@@ -1737,8 +1737,9 @@ public class ProjectionOperation implements FieldsExposingAggregationOperation {
 		public Document toDocument(AggregationOperationContext context) {
 
 			Document projections = new Document();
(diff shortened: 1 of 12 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/ProjectionOperation.java
```diff
@@ -1700,4 +1711,35 @@ public Document toDocument(AggregationOperationContext context) {
...
+
+		/*
+		 * (non-Javadoc)
+		 * @see org.springframework.data.mongodb.core.aggregation.ProjectionOperation.Projection#toDocument(org.springframework.data.mongodb.core.aggregation.AggregationOperationContext)
+		 */
+		@Override
+		public Document toDocument(AggregationOperationContext context) {
+
+			Document projections = new Document();
+			ReflectionUtils.doWithFields(type, it -> projections.append(it.getName(), 1));
```
mp911de: We probably should differentiate here and rely on our meta-model. One might want to use property accessors instead or hand in a projection interface.

mp911de:

PR #750, opened 2019-05-15
DATAMONGO-1183 - Add support for Hashed Indexes.

christophstrobl:
...and fix field name mapping in `IndexOperations` along the way.


mp911de:
That's merged and polished.

PR #752, opened 2019-05-17
DATAMONGO-2275 - Fix NPE when mapping MongoJsonSchema used in query.

christophstrobl:
We fixed a NPE when reading raw `Document` from a collection using a query matching against a JSON schema.


mp911de:

PR #754, opened 2019-05-19
DATAMONGO-719 - Add index creation for dynamically named collections.

ngbalk:


ngbalk:
CC: @waillee @tsoberg

ngbalk:
@Waillee

nhomble:
This is awesome!

ngbalk:
@christophstrobl @odrotbohm Curious to get your feedback on this PR! Thanks!

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/index/MongoPersistentEntityIndexCreator.java
```diff
@@ -50,12 +51,13 @@
...
  * @author Christoph Strobl
  * @author Mark Paluch
+ * @author Nick Balkissoon
  */
 public class MongoPersistentEntityIndexCreator implements ApplicationListener<MappingContextEvent<?, ?>> {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(MongoPersistentEntityIndexCreator.class);
 
-	private final Map<Class<?>, Boolean> classesSeen = new ConcurrentHashMap<Class<?>, Boolean>();
+	private final Table<Class, String, Boolean> collectionsSeen = HashBasedTable.create();
```
erichhsun: `HashBasedTable` is not thread safe; you may want to consider `Tables.synchronizedTable()`: https://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Tables.html#synchronizedTable-com.google.common.collect.Table-

Also, need to be careful about what you are synchronizing on when reading and writing.

Likewise, you may want to add an additional test to verify the behavior when multiple threads are using this class. I don't have enough context to tell if this is a valid use case.

ngbalk: Great feedback - looks like spring-data-mongodb depends on an older version of guava that doesn't support the synchronizedTable() , so I went back to using a ConcurrentHashMap... curious to hear what you think.

ngbalk:
@christophstrobl  please review this PR :) 
Related ticket: https://jira.spring.io/browse/DATAMONGO-719

mp911de:
Closing as per comment in [DATAMONGO-719](https://jira.spring.io/browse/DATAMONGO-719) without merging.

PR #756, opened 2019-05-20
DATAMONGO-2067 - Allow repeated usage of CompoundIndex annotation.

christophstrobl:
includes a fix for domain type field mapping for changes stream aggregations


mp911de:

PR #764, opened 2019-06-19
DATAMONGO-2305 - Upgrade to MongoDB Java Driver 3.11.0-beta4.

christophstrobl:
Tested against _4.0.9_ and _4.2.0-rc1_ servers.
Added a delay to one of the tests where the _4.2.0-rc1_ server takes a bit longer than it predecessor when creating indexes. This can lead to `BackgroundOperationInProgressForNamespace` errors.


mp911de:
That's merged now.

PR #765, opened 2019-06-19
Improve JavaDoc of BeforeSaveCallback.

michael-simons:
Some minor typos and clarification of what actually happens.


michael-simons:
I updated my change according to the updated issue.

Now the modified instance is used after applying the callback and I also added unit tests for both `insert` and `save` scenarios.

Typos in the docs fixed nevertheless ;)

christophstrobl:
thanks @michael-simons that's been merged via 1dc6a04.

PR #766, opened 2019-06-24
DATAMONGO-2306 - Add field level encryption to JSON Schema. 

christophstrobl:
We now support encrypted fields via `MongoJsonSchema` and allow XML configuration of `com.mongodb.AutoEncryptionSettings` via a dedicated factory bean.

----

This change builds upon MongoDB Java Driver 3.11.0-beta4 that should come in via #764.


mp911de:

PR #767, opened 2019-06-25
DATAMONGO-2304 - Fix documentation anchor for entity callback API.

christophstrobl:
Depends on: spring-projects/spring-data-commons#398


mp911de:

PR #768, opened 2019-06-26
DATAMONGO-2296 - Consistent use of getCollectionName(Class) throughout MongoTemplate.

christophstrobl:


mp911de:

PR #770, opened 2019-07-02
DATAMONGO-2312 - Add support for array projections.

christophstrobl:
We now support projections that create an array via `asArray(name)` and `andArrayOf(values....)`.


mp911de:

PR #771, opened 2019-07-04
DATAMONGO-2314 - Fix query by example on nested properties.

christophstrobl:
This fix allows using `Critieria.alike` on nested properties.

```java
new Criteria("nested").alike(Example.of(probe, matching().withIgnorePaths("_class"))));
```


mp911de:

PR #773, opened 2019-07-30
DATAMONGO-2328 - Add missing target type conversions for field level type hints. 

christophstrobl:
We now support `Date -> Long`, `Date -> ObjectId`, `Script -> String` and other conversions for both reading and writing scenarios.


mp911de:
That's merged now.

PR #774, opened 2019-07-31
DATAMONGO-2327 - Add toJson method to Querydsl query support. 

christophstrobl:
This allows to obtain the raw Json representation of the query for eg. debug usage.
We also updated the `toString` method to return a full Mongo Shell compatible representation of the query including projections, order, skip and limit.

```java
where(p.lastname.eq("Matthews")).orderBy(p.firstname.asc()).offset(1).limit(5);
```
```bash
find({"lastname" : "Matthews"}).sort({"firstname" : 1}).skip(1).limit(5)
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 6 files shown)


mp911de:
That's merged now.

PR #776, opened 2019-08-01
DATAMONGO-2320 - Fix aggregation field reference for $filter operator.

christophstrobl:
We now render field and local variable references correctly when using the `$filter` aggregation operator.
Prior to this commit field references had been rendered with an additional `$` prefix.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/NestedDelegatingExpressionAggregationOperationContext.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/NestedDelegatingExpressionAggregationOperationContext.java
@@ -15,6 +15,11 @@
  */
 package org.springframework.data.mongodb.core.aggregation;
 
+import java.util.Collection;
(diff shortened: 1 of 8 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/NestedDelegatingExpressionAggregationOperationContext.java
```diff
@@ -67,7 +74,22 @@ public Document getMappedObject(Document document, Class<?> type) {
 	 */
 	@Override
 	public FieldReference getReference(Field field) {
-		return new ExpressionFieldReference(delegate.getReference(field));
+
+		FieldReference reference = delegate.getReference(field);
+		return !isInnerVariableReference(field) ? reference : new ExpressionFieldReference(delegate.getReference(field)) ;
```
EGD: maybe `new ExpressionFieldReference(reference) ;` ?

mp911de:

PR #777, opened 2019-08-09
DATAMONGO-2339 - Fix QueryMapper field name resolution for properties containing underscore.

christophstrobl:
We now prevent splitting of paths that contain underscores if the entity contains a matching property.


mp911de:

PR #778, opened 2019-08-16
DATAMONGO-2342 - Upgrade MongoDB java & reactive streams Driver to 3.11 and 1.12.

christophstrobl:
Fix a timing issue when dropping collections during tests and upgrade the ci job to MongoDB 4.2


mp911de:
That's merged now.

PR #779, opened 2019-08-20
DATAMONGO-2344 - Fix slaveOK query option not applied.

christophstrobl:
Since MongoDB 3.6 the _slaveOk_ option translates to the `ReadPreference.primaryPreferred()` that is now again applied when executing a (reactive)find operation.


mp911de:
That's merged and polished.

PR #780, opened 2019-08-22
DATAMONGO-2346 - Fix (reactive)auditing of immutable versioned entities.

christophstrobl:
We now check if the source is still the same object after potentially applying auditing modifications and make sure to pass the audited object on to the mapping layer.

Limited context scan scope of event listener tests to avoid side effects in index creation.


mp911de:
That's merged now.

PR #786, opened 2019-09-04
DATAMONGO-2357 - Fix read/write for MongoDB client.model GeoJSON types.

christophstrobl:
We now consider native GeoJSON types of the MongoDB client during conversion passing on the raw values to the driver when writing and using the configured MongoDB codecs on read.


mp911de:
That's merged now.

PR #788, opened 2019-09-12
DATAMONGO-2360 - Apply query hint to count queries.

christophstrobl:


mp911de:

PR #791, opened 2019-09-19
DATAMONGO-2374 - Fix simple type result handling for repository query methods.

christophstrobl:
should be back ported to _2.1.x_


mp911de:

PR #792, opened 2019-09-23
DATAMONGO-2377 - Fix  handling of $$value and $$this in field exposing aggregation.

christophstrobl:
Internal field references to `$$this` and `$$value` are now no longer mapped against exposed fields which had caused errors before.


mp911de:
That's merged and forward-ported now.

PR #797, opened 2019-10-17
DATAMONGO-2388 - Fix CodecConfigurationException when reading index info containing DbRef in partial filter expression.

christophstrobl:
Provide the default `CodecRegistry` when converting partial index data to its `String` representation used in `IndexInfo`.

Should be back ported to _2.2.x_ and _2.1.x_.


mp911de:

PR #798, opened 2019-10-18
DATAMONGO-2394 - Fix distance conversion for derived finder using near with GeoJson type and Distance.

christophstrobl:
GeoJson requries the distance to be in meters instead of radians, so we now make sure to convert it correctly


mp911de:

PR #802, opened 2019-10-31
DATAMONGO-2385 - Remove unnecessary null checks in MongoConverters.

christophstrobl:


mp911de:

PR #804, opened 2019-11-06
DATAMONGO-2403 - Fix aggregation simple type result retrieval from empty document.

christophstrobl:
Projections used within an aggregation pipeline can result in empty documents emitted by the driver. We now guarded those cases and skip these documents within a `Flux` or simply return an empty `Mono` depending on the methods signature.


mp911de:

PR #805, opened 2019-11-08
DATAMONGO-2409 - Fix return type of Kotlin extension function for ReactiveFindOperation.DistinctWithProjection.asType().

christophstrobl:
Should be back ported to _2.2.x_ and _2.1.x_


mp911de:

PR #808, opened 2019-11-11
DATAMONGO-2237 - Do not flatten _id in aggregation results

mp911de:
We now no longer flatten the `_id` field in aggregations when using the imperative template API.

Omitting flattening allows result consumption using the same structure that was used to read and write the underlying MongoDB document.

Flattening was only applied in the imperative API but not the reactive one so both APIs behave now consistently.

---

Related ticket: [DATAMONGO-2237](https://jira.spring.io/browse/DATAMONGO-2237).


mp911de:
Will be fixed with #821 so closing this duplicate.

PR #810, opened 2019-11-26
DATAMONGO-2400 - Consider java.time.Instant a store supported native type and add configuration options for other java.time types.

christophstrobl:
We now use the MongoDB Java driver `InstantCodec` instead of a dedicated converter. Other `java.time types` like `LocalDate` use a different zone offset when writing values which can lead to unexpected behavior. Therefore we added configuration options to `MongoCustomConversions` that allow to tell the conversion sub system which approach to use when writing those kind of types.

Depends on: spring-projects/spring-data-commons#421

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MongoCustomConversions.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MongoCustomConversions.java
@@ -15,26 +15,42 @@
  */
 package org.springframework.data.mongodb.core.convert;
 
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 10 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MongoCustomConversions.java
```diff
@@ -99,4 +138,158 @@ public Object convert(@Nullable Object source, TypeDescriptor sourceType, TypeDe
...
+		 * {@link org.bson.codecs.Codec} than the Spring Data {@link Converter}.
+		 */
+		private static final List<Class<?>> JAVA_DRIVER_TIME_SIMPLE_TYPES = Arrays.asList(LocalDate.class, LocalTime.class,
+				LocalDateTime.class);
+
+		private boolean useNativeDriverJavaTimeCodecs = false;
+		private List<Object> customConverters = new ArrayList<>();
+
+		/**
+		 * Set wether or not to use the native MongoDB Java Driver {@link org.bson.codecs.Codec codes} for
```
martin-g: s/wether/whether/

mp911de:

PR #814, opened 2019-12-04
DATAMONGO-2418 - Obtain EvaluationContextExtensions lazily when parsing Bson queries.

christophstrobl:
This fix puts a patch on the issue when an annotated query does not require a SpEL context. However it does not solve the eager resolution of potential root objects in `ExtensionAwareEvaluationContextProvider` (see: [DATACMNS-1625](https://jira.spring.io/browse/DATACMNS-1625)) and therefore is likely to fail if the actual query uses an expression.

Should be forward ported to 2.3 / 3.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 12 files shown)


mp911de:

PR #817, opened 2020-01-07
DATAMONGO-2440 - Fix field target type conversion for collection values. 

christophstrobl:
We now preserve the collection nature of the source type when applying custom target type conversions. Prior to this change collection values had been changed to single element causing query errors in MongoDB when using $in queries.


mp911de:

PR #818, opened 2020-01-08
DATAMONGO-2442 - Fix $cond builder thenValueOf. 

christophstrobl:
We now create a field reference when calling the builder instead of using the value as is.


mp911de:

PR #820, opened 2020-01-14
DATAMONGO-2450 - Apply array filters to bulk write.

christophstrobl:


mp911de:

PR #821, opened 2020-01-14
DATAMONGO-2437 - Fix complex id handling when reading aggregation result.

christophstrobl:
Removed unwrapping complex id values.
`Template.aggregate` now behaves like `aggregateStream` or its reactive counterpart.


mp911de:

PR #822, opened 2020-01-15
DATAMONGO-2451 - Fix value conversion for id properties used in sort expression.

christophstrobl:
Previously we falsely converted the sort value (1/-1) into the id types target value when a `@Field` annotation had been present.


mp911de:

PR #825, opened 2020-01-17
DATAMONGO-2406 - Derived reactive deleteBy query execution should allow Mono<Void> result.

christophstrobl:
`Mono<Void>` is now a supported return type of derived reactive _deleteBy_ queries like:

```java
Mono<Void> deleteByLastname(String lastname);
```


mp911de:

PR #826, opened 2020-01-17
DATAMONGO-1997 - Add support to return the single deleted item for a deleteBy query method.

christophstrobl:
Added support for:

```java
@Nullable
Person deleteSingleByLastname(String lastname);

Optional<Person> deleteByBirthdate(Date birthdate);

Mono<Person> deleteSinglePersonByLastname(String lastname);
```


mp911de:

PR #827, opened 2020-01-21
DATAMONGO-2449 - Evaluate allowDiskUse added to Meta annotation when executing derived Aggregation.

christophstrobl:
In some scenarios aggregations might require additional options like a maximum execution time, additional log comments or the permission to temporarily write data to disk.
Use the `@Meta` annotation to set those options via `maxExecutionTimeMs`, `comment` or `allowDiskUse`.

```java
public interface PersonRepository extends CrudReppsitory<Person, String> {

  @Meta(allowDiskUse = true)
  @Aggregation("{ $group: { _id : $lastname, names : { $addToSet : $firstname } } }")
  List<PersonAggregate> groupByLastnameAndFirstnames();
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.0.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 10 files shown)


mp911de:
That's merged now.

PR #828, opened 2020-01-23
DATAMONO-2365 - Pass on index name to query hint.

christophstrobl:
along the lines extract some common functionality to centralized component.


mp911de:

PR #829, opened 2020-01-27
DATAMONGO-2457, DATAMONGO-2457 - Documentation update.

christophstrobl:
Fix id type explanation in repository documentation.
Add sample for passing on limit and offset using reactive repositories.


mp911de:

PR #830, opened 2020-01-28
DATAMONGO-2460 - Fix target type computation for complex id properties with @Field annotation.

christophstrobl:
We now set the target type to `org.bson.Document` for _id_ properties annotated with `@Field` having the implicit target type derived from the annotation. 
Along the lines we fixed warning when an _id_ property with explicit (unsupported) field name is detected.


mp911de:

PR #832, opened 2020-02-04
DATAMONGO-2079 - MappingMongoConverter no longer implements ValueResolver

mp911de:
`MappingMongoConverter` no longer implements a package-private interface so that converter instances can be proxied.

---

Related ticket: [DATAMONGO-2079](https://jira.spring.io/browse/DATAMONGO-2079).

Should be backported to 2.1.x and 2.2.x. 6946c24 should be merged only for 3.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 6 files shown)


odrotbohm:
Nice!

odrotbohm:
Looks generally good to me but don't want to get into the way of a final vote by @christophstrobl.

PR #833, opened 2020-02-10
DATAMONGO-2341 - Support shard key in template.save via Sharded annotation.

christophstrobl:
Spring Data MongoDB uses the `@Sharded` annotation to identify entities stored in sharded collections.
The shard key consists of a single or multiple properties present in every document within the target collection, and is used to distribute them across shards.

Spring Data MongoDB will do best effort optimisations for sharded scenarios when using repositories by adding required shard key information, if not already present, to `replaceOne` filter queries when upserting entities. This may require an additional server round trip to determine the actual value of the current shard key.

By setting `@Sharded(immutableKey = true)` no attempt will be made to check if an entities shard key changed.

Please see the MongoDB Documentation for further details and the list below for which operations are eligible to auto include the shard key.

* `Reactive/CrudRepository.save(...)` 
* `Reactive/CrudRepository.saveAll(...)`
* `Reactive/MongoTemplate.save(...)`


mp911de:

PR #835, opened 2020-02-18
DATAMONGO-2476 - Fix Json parsing for unquoted placeholders in arrays.

christophstrobl:


mp911de:

PR #836, opened 2020-02-19
DATAMONGO-2478 - Fix NPE in Query.of when given a proxied source.

christophstrobl:


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Query.java
```diff
@@ -517,13 +518,13 @@ public boolean isSorted() {
...
-		target.criteria.putAll(source.criteria);
-		target.skip = source.skip;
-		target.limit = source.limit;
-		target.sort = Sort.unsorted().and(source.sort);
-		target.hint = source.hint;
-		target.collation = source.collation;
-		target.restrictedTypes.addAll(source.restrictedTypes);
+		Query theQuery = ProxyUtils.unwrapTargetSource(source);
+
+		target.skip = theQuery.skip;
```
mp911de: How about using accessor methods instead of field access? That way, we can entirely omit any kind of proxy handling.

christophstrobl: good point :)

mp911de:

PR #837, opened 2020-02-22
DATAMONGO-2479 - Add AfterConvertCallback and ReactiveAfterConvertCal…

rpuch:
…lback.

Currently, only BeforeConvertCallback and BeforeSaveCallback are supported (and their reactive counterparts). This PR adds support for 'after-convert' event using entity callbacks feature.

Please note that I changed ReactiveMongoTemplate.DocumentCallback interface making its only method reactive. It seems ok because it is only used inside reactive contexts (actually, it was only called when doing a `map()` on a `Flux`/`Mono`, so now it is used in `concatMap()`/`flatMap()`).

I've only implemented one callback type ('after-convert') for now because I'd like to get an opinion from project member(s) on this issue: am I on the right track, or maybe it should not be implemented at all. After all, the original authors only implemented 2 callbacks for some reason.

What do you think?


rpuch:
There is a newer PR that also implements 'after-save' callback https://github.com/spring-projects/spring-data-mongodb/pull/839, so I'm closing this one.

PR #838, opened 2020-02-27
DATAMONGO-2481 - Improve build time.

christophstrobl:
Introduce Junit Jupiter extensions.
Flush collections instead of dropping them.
Apply cursor timeout to change stream and capped collections in tests.


mp911de:

PR #839, opened 2020-03-01
DATAMONGO-2479 - Add AfterConvertCallback and AfterSaveCallback (and …

rpuch:
…their reactive versions).

Previously, only BeforeConvertCallback and BeforeSaveCallback were supported (and their reactive counterparts). This commit adds support for 'after-convert' and 'after-save' events using entity callbacks feature.

Please note that I changed `ReactiveMongoTemplate.DocumentCallback` interface making its only method reactive. It seems ok because it is only used inside reactive contexts (actually, it was only called when doing a `map()` on a `Flux`/`Mono`, so now it is used in `concatMap()`/`flatMap()`).

This supersedes https://github.com/spring-projects/spring-data-mongodb/pull/837 , so I am closing that.

What do you think?

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/AfterSaveCallback.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/AfterSaveCallback.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.mongodb.core.mapping.event;
+
+import org.bson.Document;
+import org.springframework.data.mapping.callback.EntityCallback;
+
+/**
+ * Entity callback triggered after save of a document.
+ *
+ * @author Roman Puchkovskiy
+ * @since 3.0
+ */
+@FunctionalInterface
+public interface AfterSaveCallback<T> extends EntityCallback<T> {
+
+	/**
+	 * Entity callback method invoked after a domain object is saved. Can return either the same or a modified instance
+	 * of the domain object.
+	 *
+	 * @param entity the domain object that was saved.
+	 * @param document {@link Document} representing the {@code entity}.
+	 * @param collection name of the collection.
+	 * @return the domain object that was persisted.
+	 */
+	T onAfterSave(T entity, Document document, String collection);
+}
--- a/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoTemplateUnitTests.java
+++ b/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoTemplateUnitTests.java
@@ -25,7 +25,9 @@ import java.math.BigInteger;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -37,6 +39,7 @@ import org.assertj.core.api.Assertions;
 import org.bson.Document;
 import org.bson.conversions.Bson;
 import org.bson.types.ObjectId;
+import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
@@ -81,6 +84,9 @@ import org.springframework.data.mongodb.core.index.MongoPersistentEntityIndexCre
 import org.springframework.data.mongodb.core.mapping.Field;
 import org.springframework.data.mongodb.core.mapping.MongoMappingContext;
 import org.springframework.data.mongodb.core.mapping.event.AbstractMongoEventListener;
+import org.springframework.data.mongodb.core.mapping.event.AfterConvertCallback;
+import org.springframework.data.mongodb.core.mapping.event.AfterSaveCallback;
+import org.springframework.data.mongodb.core.mapping.event.AfterSaveEvent;
 import org.springframework.data.mongodb.core.mapping.event.BeforeConvertCallback;
 import org.springframework.data.mongodb.core.mapping.event.BeforeConvertEvent;
 import org.springframework.data.mongodb.core.mapping.event.BeforeSaveCallback;
@@ -97,10 +103,13 @@ import org.springframework.lang.Nullable;
 import org.springframework.test.util.ReflectionTestUtils;
 import org.springframework.util.CollectionUtils;
 
+import com.google.common.collect.ImmutableMap;
 import com.mongodb.MongoClientSettings;
 import com.mongodb.MongoException;
 import com.mongodb.MongoNamespace;
 import com.mongodb.ReadPreference;
+import com.mongodb.ServerAddress;
+import com.mongodb.ServerCursor;
 import com.mongodb.WriteConcern;
 import com.mongodb.client.AggregateIterable;
 import com.mongodb.client.DistinctIterable;
@@ -129,6 +138,7 @@ import com.mongodb.client.result.UpdateResult;
  * @author Christoph Strobl
  * @author Mark Paluch
  * @author Michael J. Simons
+ * @author Roman Puchkovskiy
  */
 @MockitoSettings(strictness = Strictness.LENIENT)
 public class MongoTemplateUnitTests extends MongoOperationsUnitTests {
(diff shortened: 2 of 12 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/AfterSaveCallback.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.springframework.data.mongodb.core.mapping.event;
+
+import org.bson.Document;
+import org.springframework.data.mapping.callback.EntityCallback;
+
+/**
+ * Entity callback triggered after save of a document.
+ *
+ * @author Mark Paluch
```
mp911de: Feel free to add only your name here, since this is a new file. No need to retain the authors from other types.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoTemplateUnitTests.java
```diff
@@ -1921,6 +1931,243 @@ void saveVersionedShouldProjectOnShardKeyWhenLoadingExistingDocument() {
...
+
+		Document document = initialLukeDocument();
+		when(findIterable.iterator()).thenReturn(new OneElementCursor<>(document));
+
+		template.find(new Query(), Person.class);
+
+		verify(afterConvertCallback).onAfterConvert(eq(initialLuke()), eq(document), anyString());
+	}
+
+	@NotNull
```
mp911de: Nit: No need for `@NotNull`.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoTemplateUnitTests.java
```diff
@@ -1921,6 +1931,243 @@ void saveVersionedShouldProjectOnShardKeyWhenLoadingExistingDocument() {
...
+	private Document initialLukeDocument() {
+		return new Document(ImmutableMap.of(
+				"_id", "init",
+				"firstname", "luke"
+		));
+	}
+
+	@NotNull
```
mp911de: Nit: No need for `@NotNull`.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/ReactiveMongoTemplateUnitTests.java
```diff
@@ -1096,6 +1109,282 @@ void saveVersionedShouldProjectOnShardKeyWhenLoadingExistingDocument() {
...
+
+		List<Person> results = template.find(new Query(), Person.class).timeout(Duration.ofSeconds(1))
+				.toStream().collect(Collectors.toList());
+
+		verify(afterConvertCallback).onAfterConvert(eq(initialLuke()), eq(document), anyString());
+		assertThat(results.get(0).id).isEqualTo("after-convert");
+	}
+
+	@NotNull
+	private Document initialLukeDocument() {
```
mp911de: Nit: No need for `@NotNull`.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/ReactiveMongoTemplateUnitTests.java
```diff
@@ -1096,6 +1109,282 @@ void saveVersionedShouldProjectOnShardKeyWhenLoadingExistingDocument() {
...
+		return new Document(ImmutableMap.of(
+				"_id", "init",
+				"firstname", "luke"
+		));
+	}
+
+	@NotNull
```
mp911de: Nit: No need for `@NotNull`.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/UncustomizableFindPublisher.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * {@link FindPublisher} implementation which does nothing in all FindPublisher's eigen methods, but delegates
+ * {@link #subscribe(Subscriber)} call to a {@link Publisher} passed in the constructor.
+ *
+ * @author Roman Puchkovskiy
+ */
+@RequiredArgsConstructor
+class UncustomizableFindPublisher<T> implements FindPublisher<T> {
```
mp911de: Doesn't it rather make sense to use a proper Mockito mock instead of introducing a `FindPublisher` implementation? This approach requires us to catch up each time MongoDB driver introduces any changes to `FindPublisher`.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/AfterSaveCallback.java
rpuch: Fixed

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoTemplateUnitTests.java
rpuch: Removed the annotation

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoTemplateUnitTests.java
rpuch: Removed the annotation

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/ReactiveMongoTemplateUnitTests.java
rpuch: Removed the annotation

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/ReactiveMongoTemplateUnitTests.java
rpuch: Removed the annotation

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/UncustomizableFindPublisher.java
rpuch: To be honest, with the number of methods there are on this interface, I thought it would be a nightmare to mock them all. But it looks like all the needed methods are already mocked. I've applied your advice and it seems working ok still being succinct.

rpuch:
Thank you for your review, @mp911de ! I've addressed all the issues you mentioned (if I don't miss anything).

As for the performance-related investigation: could you please give me a hint where should I look if you, by chance, already have such a tool (like a benchmark) in the code?

mp911de:
We have a few benchmarks here: https://github.com/spring-projects/spring-data-dev-tools/tree/master/benchmark/mongodb

Feel free to have a lot.

rpuch:
I've played wth the benchmarks a bit.

First, I had to migrate the benchmarks project to compile with spring-boot 2.3.0 dependencies. I will reference the PR with changes a bit later.

I ran two benchmark sessions. First is against commit 7e3f7bd86137b0d9bd1d4cb3c09ba19bb7674723 on master of spring-data-mongodb (that's the commit to which my commit in branch DATAMONGO-2479 corresponds), so this was the baseline. I'm calling this session 'master', and the corresponding archive is master.tar.gz.

Second session was against commit a54ef345c6547d503fde58070375c004e82d2204 in branch DATAMONGO-2479, I call this session 'branch', the corresponding archive is branch.tar.gz.

From what I see, the results are pretty much the same. For example, here is what I got on master:

```
Benchmark                                        Mode  Cnt        Score       Error  Units
MongoDbBenchmark.convertSingleBook              thrpt   10  2031772,058 ± 27039,347  ops/s
MongoDbBenchmark.findAll                        thrpt   10    13587,815 ±  1000,794  ops/s
MongoDbBenchmark.findByTitle                    thrpt   10    13846,553 ±   757,232  ops/s
MongoDbBenchmark.findByTitleOptional            thrpt   10    12856,971 ±  2289,165  ops/s
MongoDbBenchmark.rawFindAll                     thrpt   10    15628,538 ±  1230,409  ops/s
MongoDbBenchmark.rawFindByTitle                 thrpt   10    15742,024 ±   412,087  ops/s
MongoDbBenchmark.repositoryFindAll              thrpt   10    10783,801 ±  3571,915  ops/s
MongoDbBenchmark.repositoryFindByTitle          thrpt   10    12384,333 ±   799,111  ops/s
MongoDbBenchmark.repositoryFindByTitleDeclared  thrpt   10    13278,882 ±   785,733  ops/s
MongoDbBenchmark.repositoryFindByTitleOptional  thrpt   10     9310,293 ±  3023,870  ops/s
```

and here are the results on the branch:

```
Benchmark                                        Mode  Cnt        Score       Error  Units
MongoDbBenchmark.convertSingleBook              thrpt   10  2045614,403 ± 40567,310  ops/s
MongoDbBenchmark.findAll                        thrpt   10    10952,504 ±   234,477  ops/s
MongoDbBenchmark.findByTitle                    thrpt   10    11276,996 ±  1772,170  ops/s
MongoDbBenchmark.findByTitleOptional            thrpt   10    11642,227 ±  1720,294  ops/s
MongoDbBenchmark.rawFindAll                     thrpt   10    14637,392 ±  1520,145  ops/s
MongoDbBenchmark.rawFindByTitle                 thrpt   10    15816,345 ±  1077,191  ops/s
MongoDbBenchmark.repositoryFindAll              thrpt   10     9705,944 ±   292,123  ops/s
MongoDbBenchmark.repositoryFindByTitle          thrpt   10    10444,036 ±  2058,041  ops/s
MongoDbBenchmark.repositoryFindByTitleDeclared  thrpt   10    13733,597 ±   758,627  ops/s
MongoDbBenchmark.repositoryFindByTitleOptional  thrpt   10    11619,635 ±   773,341  ops/s
```

 I'm attaching the full results (logs and reports) as archives.

[master.tar.gz](https://github.com/spring-projects/spring-data-mongodb/files/4332661/master.tar.gz)
[branch.tar.gz](https://github.com/spring-projects/spring-data-mongodb/files/4332662/branch.tar.gz)

I also added one more benchmark: `AfterConvertCallbacksBenchmark`. It is modeled after `CallbacksBenchmark`.

mp911de:
Great work, thank you. There's a slight difference if you check `findAll` (`13587,815` vs. `10952,504`). These differences show only up when returning multiple elements. Scenarios returning just one element look unchanged.

Based on these numbers, it makes sense to merge this PR and we can investigate on our end how we can improve `EntityCallback` performance in Spring Data Commons.

rpuch:
Should I rebase on the current master?

mp911de:
Feel free, however, there's no need as GitHub does not show any merge conflicts. Please revisit the formatting changes around imports to reduce the noise in the changes. We will take care of the remaining issues during the merge.

rpuch:
Then I guess I will leave it as is if the rebase is not mandatory.
I've looked through the import changes once more, seems to be ok.

mp911de:

PR #841, opened 2020-03-11
DATAMONGO-2488 - Fix nested array path field name mapping.

christophstrobl:


mp911de:

PR #842, opened 2020-03-17
DATAMONGO-625 - Add support for GridFS with predefined id.

christophstrobl:
We now support storing GridFS content with predefined `_id`, which allows to replace an existing resource instead of having to delete and reupload it.

```java
GridFsUpload<String> upload = GridFsUpload.fromStream(resource.getInputStream())
	.id(id)
	.filename("upload.xml")
	.contentType("sync")
	.build();

gridFsTemplate.save(upload);
```

```java
ReactiveGridFsUpload<String> upload = ReactiveGridFsUpload.fromPublisher(data)
	.id(id)
	.filename("upload.xml")
	.contentType("reactive")
	.build();

reactiveGridFsTemplate.save(upload).subscribe();
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.0.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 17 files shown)


mp911de:

PR #844, opened 2020-03-19
DATAMONGO-931 - Add support for $redact aggregation. 

christophstrobl:
We now support `$redact` via `Aggregation.redact(...)`.

```java
Aggregation.redact(ConditionalOperators.when(Criteria.where("level").is(5))
    .then(RedactOperation.PRUNE)
    .otherwise(RedactOperation.DESCEND));
```


mp911de:

PR #847, opened 2020-04-01
DATAMONGO-2502 - Fix regression in field name mapping for nested array paths.

christophstrobl:
We now remove placeholders correctly which allows to traverse the path based on actual properties. This enables a correct mapping of the individual parts.

Related to [DATAMONGO-2488](https://jira.spring.io/browse/DATAMONGO-2488) (#841).


mp911de:

PR #850, opened 2020-04-07
DATAMONGO-2480 - Fix interface projection for entities with required fields that are not present in projection result.

christophstrobl:
We now use the raw result (`org.bson.Document`) for interface projections that lack required fields for constructing the actual domain type due to field limitations computed by the projection itself.


mp911de:
Closing without merge as per comment in the [associated ticket](https://jira.spring.io/browse/DATAMONGO-2480).

PR #851, opened 2020-04-07
DATAMONGO-2506 - Provide meaningful error message when using unsupported return type in repository aggregation method.

christophstrobl:
We improved the error message for unsupported return types instead of running into an `IllegalArgumentException` when trying to return a unique result.


mp911de:

PR #855, opened 2020-04-15
DATAMONGO-2513 - Fix Eq aggregation operator comparing collection values.

christophstrobl:


mp911de:

PR #856, opened 2020-04-16
DATAMONGO-2516 - Assert compatibility with MongoDB 4.4-rc0.

christophstrobl:
Fixes:
- Fields list must not contain text search score property when no `$text` criteria present.
- Sort must not be an empty document when running map reduce.
- Timeout in tests creating indexes.

Changes:
- score property might now be `null` (was `zero` before) when not having a `$text` criteria present.


mp911de:

PR #857, opened 2020-04-17
DATAMONGO-2517 - Fix invalid entity creation for text queries.

christophstrobl:
Fix a glitch in the `MappingMongoConverter` that uses the single argument constructor (since it matches in type and parameter count to the given input string) to falsely instantiate an Entity when it should not.


mp911de:

PR #859, opened 2020-04-21
DATAMONGO-2523 - Fix Json binding of SpEL expressions in arrays.

christophstrobl:
The closing bracket must not have a leading whitespace.


mp911de:

PR #863, opened 2020-05-12
DATAMONGO-2542 - Shortcut PersistentPropertyPath resolution during query creation.

christophstrobl:
By shortcutting the path resolution we avoid checking keywords like `$in` against a potential path expression.


mp911de:

PR #864, opened 2020-05-13
DATAMONGO-2545 - Fix regression in String query SpEL parameter binding.

christophstrobl:
The PR includes a fix for parameter binding within a SpEL expression using query parameter placeholders `?0`, `?1`,... instead of the array index `[0]`,`[1]`,...

```java
@Query("{ 'hidden' : ?#{ hasRole('batman') ? '?0' : '?1' } }"
List<User> findAllUsingSpelBy(String arg0, String arg1);
```

It also fixes the case where the SpEL expression actually results in the query document to use.

```java
@Query("?#{ hasRole('admin') ? { 'region': 'all' } : { 'user.superviser':  principal.id } }")
List<User> findAllUsingSpelBy();
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.1.0-SNAPSHOT</version>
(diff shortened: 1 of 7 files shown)


mp911de:

PR #865, opened 2020-05-14
DATAMONGO-2547 - Use target class ClassLoader instead of default CL when creating proxy instances.

christophstrobl:
related to: spring-projects/spring-boot#21438


mp911de:

PR #867, opened 2020-05-26
DATAMONGO-2556 - Add estimatedCount for collections.

christophstrobl:
The newly introduced methods delegate to the drivers `MongoCollection.estimatedDocumentCount`.


mp911de:
That's merged now.

PR #871, opened 2020-06-18
DATAMONGO-2570 - Change type of count variable to primitive.

mehranus:
For avoiding unintentional unboxing I simply change the `count` variable to the primitive type of `long`.


mp911de:

PR #875, opened 2020-07-03
DATAMONGO-2490 - Fix dbref fetching during ongoing transaction.

christophstrobl:


mp911de:

PR #876, opened 2020-07-14
DATAMONGO-2536 - Add option to skip reading aggregation result.

christophstrobl:


mp911de:

PR #879, opened 2020-07-23
DATAMONGO-2557 - Use configured CodecRegistry when parsing String based queries instead of default one.

christophstrobl:


mp911de:

PR #881, opened 2020-08-08
Fix Document example, Aggregation input parameter syntax error

kioyong:
should be ?0 instead of $?0, Otherwise, an exception will be thrown
"JSON reader was expecting ':' but found '}'."

- [  ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATAMONGO).
- [  ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [  ] You submit test cases (unit or integration tests) that back your changes.
- [  ] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).


christophstrobl:
Thanks @kioyong! The PR will be part of [DATAMONGO-2638](https://jira.spring.io/browse/DATAMONGO-2638).

christophstrobl:
Merged to _main_ line and back ported to maintenance branches.

PR #886, opened 2020-09-17
DATAMONGO-2622 - Add support for $unionWith aggregation stage. 

christophstrobl:
We now support the `$unionWith` aggregation stage via the `UnionWithOperation` that performs a union of two collections by combining pipeline results, potentially containing duplicates, into a single result set that is handed over to the next stage.
In order to remove duplicates it is possible to append a `GroupOperation` right after `UnionWithOperation`.
If the `UnionWithOperation` uses a pipeline to process documents, field names within the pipeline will be treated as is. In order to map domain type property names to actual field names (considering potential `@Field` annotations) make sure the enclosing aggregation is a `TypedAggregation` and provide the target type for the `$unionWith` stage via `mapFieldsTo(Class)`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.1.0-SNAPSHOT</version>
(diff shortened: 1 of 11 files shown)


mp911de:

PR #887, opened 2020-09-28
DATAMONGO-2623 - Add support for $function & $accumulator aggregation operators.

christophstrobl:


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/ScriptOperators.java
```diff
@@ -0,0 +1,412 @@
...
+		private Accumulator(Map<String, Object> value) {
+			super(value);
+		}
+
+		/**
+		 * Create a new {@link Accumulator}.
+		 *
+		 * @return new instance of {@link Accumulator}.
+		 */
+		public static Accumulator accumulator() {
```
mp911de, line 211: Would it make sense to have `Accumulator` being used in a builder style? Calling `Accumulator.accumulator()` basically creates a ready-to-use `AggregationExpression` without having values for all fields.

When used fluently, this doesn't make a difference, but having an object that pretends to be ready to use opens up for wrong usage.

mp911de:

PR #889, opened 2020-10-06
DATAMONGO-2616 - Short circuit id value conversion in MongoConverter.

christophstrobl:


mp911de:

PR #890, opened 2020-11-09
DATAMONGO-2635 - Enforce aggregation pipeline mapping.

christophstrobl:
Avoid using the `Aggregation.DEFAULT_CONTEXT` which does not map contained values to the according MongoDB representation. We now use a relaxed aggregation context, preserving given field names, where possible.
Along the lines also change the behaviour of `_id` field projections to allow the inclusion of the filed nevertheless the current `AggregationOperationContext` is aware of it.


mp911de:

PR #892, opened 2020-11-13
DATAMONGO-2652 - Implements deleteAllById.

schauder:
Merge after DATACMNS-800

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleMongoRepository.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleMongoRepository.java
@@ -51,6 +51,7 @@ import com.mongodb.client.result.DeleteResult;
  * @author Thomas Darimont
  * @author Mark Paluch
  * @author Mehran Behnam
+ * @author Jens Schauder
  */
 public class SimpleMongoRepository<T, ID> implements MongoRepository<T, ID> {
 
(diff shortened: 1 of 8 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleReactiveMongoRepository.java
```diff
@@ -404,6 +405,14 @@ public SimpleReactiveMongoRepository(MongoEntityInformation<T, ID> entityInforma
 		return Flux.fromIterable(entities).flatMap(this::delete).then();
 	}
 
+	@Override
+	public Mono<Void> deleteAllById(Iterable<? extends ID> ids) {
+
+		Assert.notNull(ids, "The given Iterable of ids must not be null!");
+
+		return Flux.fromIterable(ids).flatMap(this::deleteById).then();
```
mp911de, line 413: We should collect all ID's into a query to run a single delete query and to avoid issuing a query per given Id.

mp911de (changes requested):

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleReactiveMongoRepository.java
schauder: Should we do the same for `public Mono<Void> deleteAll(Iterable<? extends T> entities)`?

schauder: I made the same change for `public Mono<Void> deleteAll(Iterable<? extends T> entities)` as well in a separate commit. 

Pick and choose.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleMongoRepository.java
```diff
@@ -182,11 +183,19 @@ public void delete(T entity) {
...
 
 		entities.forEach(this::delete);
 	}
 
+	@Override
+	public void deleteAllById(Iterable<? extends ID> ids) {
+
+		Assert.notNull(ids, "The given Iterable of ids must not be null!");
+
+		ids.forEach(this::deleteById);
```
mp911de, line 196: We should use a single query here. Going to apply that change during the merge.

mp911de (approved):

mp911de:

PR #896, opened 2020-12-15
DATAMONGO-1902 - Add support for embedding objects.

christophstrobl:
**Depends on:** spring-projects/spring-data-commons#2293

⚠️ Open for discussion -   do not merge! 

----

### Embedded Types

Embedded entities are used to design value objects in your Java domain model whose properties are flattened out into the MongoDB Document.

#### Embedded Types Mapping

In the example below you see, that `User.name` is annotated with `@Embedded`.
The consequence of this is that all properties of `UserName` are folded into the `user` document.

```java
public class User {
    @Id
    private String userId;

    @Embedded(onEmpty = USE_NULL)
    private UserName name;
}

public class UserName {
    private String firstname;
    private String lastname;
}
```

```json
{
  "_id" : "1da2ba06-3ba7",
  "firstname" : "Emma",
  "lastname" : "Frost"
}
```

It is possible to use complex types within an embedded object.
However those must not be, nor contain embedded fields themselves.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
+	<version>3.2.0-DATAMONGO-1902-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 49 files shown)


mp911de:
In case we don't feel comfortable merging the related PR in Spring Data Commons (spring-projects/spring-data-commons#480), let's move forward by introducing a store-specific embedded-annotation and sort out the duplications in `PersistentEntity`/`PersistentProperty` at a later stage.

mp911de:

PR #897, opened 2020-12-16
DATAMONGO-2671 - Fix dateFromParts millisecond field name. 

christophstrobl:
Use `millisecond` instead of `milliseconds` field name.

Related to: https://jira.mongodb.org/browse/DOCS-10652


mp911de:
Fixes #3523.

PR #3541, opened 2021-01-21
Issue/3.1.x/3055

christophstrobl:
Update `count` vs. `estimatedCount` documentation and allow access to `mongoDatabaseFactory` used in `ReactiveMongoTemplate` so users can subclass the template and override `doCount` while being able to check for the presence of an ongoing transaction using `ReactiveMongoDatabaseUtils.isTransactionActive(getDatabaseFactory())`.


mp911de:

PR #3545, opened 2021-01-25
Use relaxed type mapping for aggregations by default.

christophstrobl:
Switch from a _strict_ to a _relaxed_ type mapping for aggregation executions. 
This allows users to add fields to the `Aggregation` that might be part of the stored document but not necessarily of its java model representation.
Instead of throwing an `Exception` the relaxed type check will go on with the user provided field names.
To restore the original behaviour use the `strictMapping` option on `AggregationOptions`.


mp911de:

Issue #3546, opened 2021-01-25
org.bson.codecs.configuration.CodecConfigurationException: The uuidRepresentation has not been specified, so the UUID cannot be encoded.


airtaste:
I'm trying to upgrade the version of spring boot from 2.3.0 to 2.4.0. After updates I've tried to run my unit tests and got codec exception :disappointed: 

> Caused by: org.bson.codecs.configuration.CodecConfigurationException: The uuidRepresentation has not been specified, so the UUID cannot be encoded.

	at org.bson.codecs.UuidCodec.encode(UuidCodec.java:72)
	at org.bson.codecs.UuidCodec.encode(UuidCodec.java:37)
	at org.bson.codecs.EncoderContext.encodeWithChildContext(EncoderContext.java:91)
	at org.bson.codecs.DocumentCodec.writeValue(DocumentCodec.java:203)
	at org.bson.codecs.DocumentCodec.writeMap(DocumentCodec.java:217)
	at org.bson.codecs.DocumentCodec.encode(DocumentCodec.java:159)
	at org.bson.codecs.DocumentCodec.encode(DocumentCodec.java:46)
	at org.bson.Document.toJson(Document.java:440)
	at org.bson.Document.toJson(Document.java:413)
	at org.bson.Document.toJson(Document.java:400)
	at org.springframework.data.mongodb.core.convert.MongoConverters$DocumentToStringConverter.convert(MongoConverters.java:241)
	at org.springframework.data.mongodb.core.convert.MongoConverters$DocumentToStringConverter.convert(MongoConverters.java:234)
	at org.springframework.core.convert.support.GenericConversionService$ConverterAdapter.convert(GenericConversionService.java:386)
	at org.springframework.core.convert.support.ConversionUtils.invokeConverter(ConversionUtils.java:41)

An example of a simplified project: https://github.com/airtaste/example/tree/master

Also, I've tried to use Spring Boot property spring.data.mongodb.uuid-representation and set it to 'standard', but the error still occurred. Furthermore, I tried to customize MongoClientSettings to include UuidRepresentation.STANDARD codec registry. Ineffectually.

It's a very odd issue.

Thanks in advance!

christophstrobl:
Having a look at the example the error in `LessonActivityRepositoryUnitTest` results from the log statement invoking `org.bson.Document#toJson()` using the default `Encoder` that does not have the UUID setting. 
[...]

PR #3551, opened 2021-02-08
Use UUID toString representation for UUID values when converting org.bson.Document to json String

christophstrobl:
This PR switches the rendering of `UUID` values to their `toString` format when converting `org.bson.Document` to json via the `DocumentToStringConverter`.

This will move the resulting representation from `{"$binary": "QUK3ZihZ9cdhWjTf5TZqrw==", "$type": "03"}` 
to `480971b0-7160-4120-acd0-6fd6b82418ad`.

The conversion only applies on read in cases where an entire `Document` (like a composite id), is mapped to a `String` property of the domain model. 
Obviously not the most common scenario but one that might easily happen during projections or aggregations.

Relates to: #3546

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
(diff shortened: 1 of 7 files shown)


mp911de:
That's merged now. The UUID fix is backported to 3.1.x.

PR #3554, opened 2021-02-16
Fix case insensitive derived in queries on String properties.

christophstrobl:
We now consider the `IgnoreCase` part of a derived query when used along with `In`.
```java
repository.findByLastnameIgnoreCaseIn("bEAuFoRd", "maTTheWs")
```

Closes: #3395


mp911de:

Issue #3552, opened 2021-02-12
UpdateMapper drops numeric keys in Maps


RamThirupathy:
Numeric Bson Ids are dropped when using findAndModify method in org.springframework.data.mongodb.core.MongoTemplate.

**Ex:**
```
public class Foo {
 private Map<String, String> map;
 
 private Map<String, String> getMap() {
   return map;
 }

 private void setMap(Map<String, String> map) {
   this.map = map
  }
}
```

```
@Test
public void test() {
    Foo foo = new Foo();
    mongoTemplate.save(foo);
    Update update = new Update().set("map.601218778970110001827396", "testing");
    druryIssue = mongoTemplate.findAndModify(new Query(Criteria.where("_id").is(foo.getId())), update,
          new FindAndModifyOptions().returnNew(true),  Foo.class);
    System.out.println(druryIssue.map);
}
```

Foo is saved us

```
{
    "_id" : ObjectId("6023ff93defc2460cc3c4a26"),
    "_class" : "com.test.Foo",
    "map" : "testing"//it should have been "map.601218778970110001827396":"testing"
}

```

and resulting in exception
```

org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [java.lang.String] to type [java.util.Map<?, ?>]
	at org.springframework.core.convert.support.GenericConversionService.handleConverterNotFound(GenericConversionService.java:324)
	at org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:206)
	at org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:187)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.getPotentiallyConvertedSimpleRead(MappingMongoConverter.java:838)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readValue(MappingMongoConverter.java:1230)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.access$200(MappingMongoConverter.java:84)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$MongoDbPropertyValueProvider.getPropertyValue(MappingMongoConverter.java:1176)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.getValueInternal(MappingMongoConverter.java:886)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$1.doWithPersistentProperty(MappingMongoConverter.java:290)
[...]

PR #3565, opened 2021-02-19
Preserve numeric keys that exceed Long range when mapping Updates.

christophstrobl:
This PR makes sure we preserve numeric map keys no matter what.

Closes: #3552


mp911de:

PR #3569, opened 2021-02-25
Translate MongoSocketException subclasses to DataAccessResourceFailureException

### LINKED ISSUE #3568: MongoSocketWriteException may be translated into DataAccessResourceFailureException
Hi

## What
This is a user feedback about spring data mongodb exception translation:
- In production we encounter a `MongoSocketWriteException` during a simple `findBy...()`
- spring-data-mongodb translates this exception into a `org.springframework.data.mongodb.UncategorizedMongoDbException`

Here is an extract of the stack:
```
org.springframework.data.mongodb.UncategorizedMongoDbException: Exception sending message; nested exception is com.mongodb.MongoSocketWriteException: Exception sending message
	at org.springframework.data.mongodb.core.MongoExceptionTranslator.translateExceptionIfPossible(MongoExceptionTranslator.java:138)
	at org.springframework.data.mongodb.core.MongoTemplate.potentiallyConvertRuntimeException(MongoTemplate.java:2793)
	at org.springframework.data.mongodb.core.MongoTemplate.executeFindMultiInternal(MongoTemplate.java:2698)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2441)
	at org.springframework.data.mongodb.core.ExecutableFindOperationSupport$ExecutableFindSupport.doFind(ExecutableFindOperationSupport.java:214)
	at org.springframework.data.mongodb.core.ExecutableFindOperationSupport$ExecutableFindSupport.all(ExecutableFindOperationSupport.java:157)
	at org.springframework.data.mongodb.repository.query.MongoQueryExecution$PagedExecution.execute(MongoQueryExecution.java:121)
	at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.execute(AbstractMongoQuery.java:99)
```

- on our side we would manage `MongoSocketException` and other "resource-related" exceptions in the same way but the translator didn't translate `MongoSocketWriteException` case into `DataAccessResourceFailureException` so we can't without checking on our side this kind of mongo exceptions..

## Version

We're using spring-data-mongodb `3.0.6.RELEASE ` 

but I saw that the relate code dont change.

List of resource-related exception:

https://github.com/spring-projects/spring-data-mongodb/blob/ad6d2c97b7b03e14fe3233509fd13588fc33db92/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java#L58-L60

## Why

Today I think the translator translates only an exact match between user-exception and `RESOURCE_FAILURE_EXCEPTIONS` translation matrix entries. 
https://github.com/spring-projects/spring-data-mongodb/blob/ad6d2c97b7b03e14fe3233509fd13588fc33db92/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java#L87

## HowToReproduce
```java
        MongoSocketWriteException writeException = new MongoSocketWriteException("oO", null, null);

        System.out.println(ClassUtils.getShortName(ClassUtils.getUserClass(writeException.getClass())));
        // MongoSocketWriteException

        System.out.println(new MongoExceptionTranslator().translateExceptionIfPossible(writeException));
        // org.springframework.data.mongodb.UncategorizedMongoDbException: oO; nested exception is com.mongodb.MongoSocketWriteException: oO
```

## Fix Suggest
There is no check on user-exception ancestor.

For example, I expect `MongoSocketWriteException` (child of  `MongoSocketException`) like all other `MongoSocketException` childs to be translated into `DataAccessResourceFailureException`. 

Am I wrong? 
If I'm right, do you plan to improve `MongoExceptionTranslator` this way ?

Regards

boly38:
Closes #3568

Improve `MongoExceptionTranslator` to accept direct children is some case. First use case : `MongoSocketException, MongoSocketException$children`.

NB/ historical contributors of this class: @christophstrobl, @michalvich, @odrotbohm (Oliver Gierke)

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.data.mongodb.core;
 
+import com.mongodb.MongoSocketException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
@@ -49,6 +50,7 @@ import com.mongodb.bulk.BulkWriteError;
  * @author Oliver Gierke
  * @author Michal Vich
  * @author Christoph Strobl
+ * @author Brice Vandeputte
  */
 public class MongoExceptionTranslator implements PersistenceExceptionTranslator {
 
@@ -78,6 +80,10 @@ public class MongoExceptionTranslator implements PersistenceExceptionTranslator 
 			throw new InvalidDataAccessApiUsageException(ex.getMessage(), ex);
 		}
 
+		if (ex instanceof MongoSocketException) {
+			return new DataAccessResourceFailureException(ex.getMessage(), ex);
+		}
+
 		String exception = ClassUtils.getShortName(ClassUtils.getUserClass(ex.getClass()));
 
 		if (DUPLICATE_KEY_EXCEPTIONS.contains(exception)) {
@@ -87,6 +93,7 @@ public class MongoExceptionTranslator implements PersistenceExceptionTranslator 
 		if (RESOURCE_FAILURE_EXCEPTIONS.contains(exception)) {
 			return new DataAccessResourceFailureException(ex.getMessage(), ex);
 		}
+
 
 		if (RESOURCE_USAGE_EXCEPTIONS.contains(exception)) {
 			return new InvalidDataAccessResourceUsageException(ex.getMessage(), ex);
--- a/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoExceptionTranslatorUnitTests.java
+++ b/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/MongoExceptionTranslatorUnitTests.java
@@ -17,7 +17,8 @@ package org.springframework.data.mongodb.core;
 
 import static org.assertj.core.api.Assertions.*;
 
-import java.net.UnknownHostException;
+import com.mongodb.MongoSocketReadTimeoutException;
+import com.mongodb.MongoSocketWriteException;
 
 import org.bson.BsonDocument;
 import org.junit.jupiter.api.BeforeEach;
@@ -45,9 +46,11 @@ import com.mongodb.ServerAddress;
  * @author Michal Vich
  * @author Oliver Gierke
  * @author Christoph Strobl
+ * @author Brice Vandeputte
  */
 public class MongoExceptionTranslatorUnitTests {
 
+	public static final String EXCEPTION_MESSAGE = "IOException";
 	MongoExceptionTranslator translator;
 
 	@BeforeEach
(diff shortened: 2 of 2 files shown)


boly38:
(sorry I'm new on this project) I'm unable to see UT+IT results from github PR checks list. Are they executed somewhere?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
```diff
@@ -78,21 +88,25 @@ public DataAccessException translateExceptionIfPossible(RuntimeException ex) {
 			throw new InvalidDataAccessApiUsageException(ex.getMessage(), ex);
 		}
 
-		String exception = ClassUtils.getShortName(ClassUtils.getUserClass(ex.getClass()));
+		Class<? extends RuntimeException> exceptionClass = ex.getClass();
```
mp911de, line 91: Let's add a specific `instanceof` check instead of introducing a pattern matching for a single case.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
```diff
@@ -49,21 +52,28 @@
...
 public class MongoExceptionTranslator implements PersistenceExceptionTranslator {
+	/**
+	 * translation matrix definitions:
+	 * - adding {@code ClassShortName} will associate a given exception class to a translated group
+	 * - adding {@code ClassShortName$children} will associate a given exception direct children to a translated group
+	 */
 
 	private static final Set<String> DUPLICATE_KEY_EXCEPTIONS = new HashSet<>(
-			Arrays.asList("MongoException.DuplicateKey", "DuplicateKeyException"));
+			asList("MongoException.DuplicateKey", "DuplicateKeyException"));
```
mp911de, line 65: Nit: We typically don't use static imports to retain the context.

mp911de (changes requested):
Thanks for your pull request. We don't have a build for pull requests. You can run the build locally if you spin up a MongoDB 4.x server locally.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
boly38: I'm confuse about your remark because I dont see how to apply:
- if I apply `instanceOf MongoSocketException` it will work for my use-case of course, 
but the question is how to not apply this for the rest (of the translation matrix entries) ?

Should we consider accepting children  for all entries ?


In my first suggest, I consider this points:
- we would like to keep a minor coupling with mongo exception hierarchy. 
- we suppose that MongoSocketException update frequency is less than their children.
- we would like to avoid listing individually each MongoSocketException children.
That's why I introduced "Class$children" a pattern matching.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
boly38: sure I would revert that on next upd

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoExceptionTranslator.java
mp911de: We already depend on `MongoException` so adding a subtype is fine. We don't have specific mappings for `MongoSocketException` so it doesn't clash with any existing mappings. Also, the error code for all `MongoSocketException`s is `-2` so its fine to follow the simplemost approach. Any further functionality isn't required right now and the easiest way is typically the one which works best.

boly38: >the easiest way is typically the one which works best.
100% agree with that, so I'll upd with a specific `MongoSocketException` case.

mp911de:

PR #3575, opened 2021-03-01
Introduce ConversionContext and clean up MappingMongoConverter

### LINKED ISSUE #3571: Introduce ConversionContext and clean up MappingMongoConverter
To simplify customization of `MappingMongoConverter`, we should introduce a strategy object that encapsulates decision paths invoking simple and container conversion. Simple conversion maps to simple types. Container conversion applies conversion for lists, maps, and entities.

We also should cleanup the converter to align method signatures of the individual converter methods, reorder methods and revisit the converter surface.

mp911de:
`ConversionContext` serves as entrpoint for recursive (read) conversion of documents, lists, maps, and simple values. The actual decision which converter strategy to apply is now encapsulated by `ConversionContext.convert(…)` which removes strategy duplications from the actual conversion methods.

Also, converter methods for documents, maps, lists, … are now `protected` for easier customization by subclasses.

Closes #3571


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/util/BsonUtils.java
```diff
@@ -62,6 +66,47 @@
...
+	}
+
+	/**
+	 * Check if a given entry (key/value pair) is present in the given {@link Bson}.
+	 *
+	 * @param bson must not be {@literal null}.
+	 * @param key must not be {@literal null}.
+	 * @param value can be {@literal null}.
+	 * @return {@literal true} if (key/value pair) is present.
+	 * @since 2.5
```
mp911de, line 91: This should be `@since 3.2`

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/util/BsonUtils.java
```diff
@@ -62,6 +66,47 @@
 		return (T) asMap(bson).get(key);
 	}
 
+	/**
+	 * Remove {@code _id : null} from the given {@link Bson} if present.
+	 * 
+	 * @param bson must not be {@literal null}.
+	 * @since 2.5
```
mp911de, line 73: This should be `@since 3.2`

PR #3581, opened 2021-03-05
Fix CustomConverter conversions.

### LINKED ISSUE #3580: Fix CustomConverter conversion lookup.
The change (be8e702) introduced a more concrete converter lookup which causes trouble using the underlying `ConversionService` that might not have the converter registered for the given type but one of its parent ones. This leads to a `ConverterNotFoundException` during the mapping process.

```java
interface Message<T> {
	T getValue();
}

class GenericMessage<T> implements Message<T> {
	T value;
}

class TypedMessage extends GenericMessage<String> {
}

@ReadingConverter
class MessageConverter implements Converter<Document, Message<?>> {
	@Override
	public Message<?> convert(org.bson.Document source) {
		// ...
	}
}

// ...

Document source = new Document("_class", TypedMessage.class.getName())
		.append("value", "v1");

Message<?> target = converter.read(Message.class, source);
```

christophstrobl:
This PR makes sure to fall back to the configured custom `Converter` if no more type specific `Converter` has been registered.

Closes #3580

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.2.0-SNAPSHOT</version>
+	<version>3.2.0-GH-3580-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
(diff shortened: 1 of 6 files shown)


artembilan (approved):
👍

Issue #3583, opened 2021-03-09
Support aggregation expression on fields projection.


christophstrobl:
MongoDB allows using aggregation expressions as part of the fields projection as shown below.

```javascript
db.getCollection('person').find({}, { "firstname" : 1, "lastname" : {"$toUpper" : "$lastname"}})
```

**Initial Support:**
* Pass on a given expression projection to the server as is.
* The `query.fields` object should be able to accept an expression eg. by using a `BasicQuery`.
* The `@Query` annotation should parse the `fields` attribute holding an expression.

**Final Support:** (might require a follow up ticket)
* Map expressions against domain types.
* Support `AggregationExpression` via `Field` projection API.

PR #3585, opened 2021-03-10
Support expressions in query field projections.

christophstrobl:
Starting with MongoDB 4.4 it is possible to use the aggregation expressions syntax for field projections as shown below.

```java
query.fields()
  // Use a native expression. The used field names must refer to the ones of the document within the database.
  .project(MongoExpression.expressionFromString("'$toUpper' : '$last_name'"))
  // Assign the field name that shall hold the expression result in the target document.
  .as("last_name"); 

query.fields()
  // AggregationExpression makes sure field names are mapped to the ones used in the domain model.
  .project(StringOperators.valueOf("lastname").toUpper()) 
  .as("last_name");

query.fields()
  // Use SpEL along with an AggregationExpression to invoke expression functions.
  .project(AggregationSpELExpression.expressionOf("toUpper(lastname)"))
  .as("last_name");
```
`@Query(fields='...')` allows usage of expression field projections at `Repository` level.

Closes: #3583

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoExpression.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoExpression.java
@@ -15,8 +15,6 @@
  */
 package org.springframework.data.mongodb;
 
-import java.util.function.Function;
-
 /**
  * Wrapper object for MongoDB expressions like {@code $toUpper : $name} that manifest as {@link org.bson.Document} when
  * passed on to the driver.
@@ -36,7 +34,7 @@ import java.util.function.Function;
  * @see org.springframework.data.mongodb.core.aggregation.DateOperators
  * @see org.springframework.data.mongodb.core.aggregation.ObjectOperators
  * @see org.springframework.data.mongodb.core.aggregation.SetOperators
- * @see org.springframework.data.mongodb.core.aggregation.StringOperators @
+ * @see org.springframework.data.mongodb.core.aggregation.StringOperators
  */
 @FunctionalInterface
 public interface MongoExpression {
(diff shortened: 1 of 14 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoExpression.java
```diff
@@ -0,0 +1,86 @@
...
+	org.bson.Document toDocument();
+
+	/**
+	 * Convert this instance to another expression applying the given conversion {@link Function}.
+	 *
+	 * @param function must not be {@literal null}.
+	 * @param <T>
+	 * @return never {@literal null}.
+	 */
+	default <T extends MongoExpression> T as(Function<MongoExpression, T> function) {
```
mp911de, line 58: how about `map`?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoExpression.java
```diff
@@ -0,0 +1,86 @@
...
+ * @since 3.2
+ * @see org.springframework.data.mongodb.core.aggregation.ArithmeticOperators
+ * @see org.springframework.data.mongodb.core.aggregation.ArrayOperators
+ * @see org.springframework.data.mongodb.core.aggregation.ComparisonOperators
+ * @see org.springframework.data.mongodb.core.aggregation.ConditionalOperators
+ * @see org.springframework.data.mongodb.core.aggregation.ConvertOperators
+ * @see org.springframework.data.mongodb.core.aggregation.DateOperators
+ * @see org.springframework.data.mongodb.core.aggregation.ObjectOperators
+ * @see org.springframework.data.mongodb.core.aggregation.SetOperators
+ * @see org.springframework.data.mongodb.core.aggregation.StringOperators @
```
mp911de, line 39: Nit: There's a lingering `@`

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoExpression.java
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * Create a new {@link MongoExpression} from plain String (eg. {@code $toUpper : $name}). <br />
+	 * The given source value will be wrapped with <code>{ }</code> to match an actual MongoDB {@link org.bson.Document}
+	 * if necessary.
+	 *
+	 * @param json must not be {@literal null}.
+	 * @return new instance of {@link MongoExpression}.
+	 */
+	static MongoExpression expressionFromString(String json) {
```
mp911de, line 70: How about `parse`?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoExpression.java
christophstrobl: Intentionally did not name this one `map`, as I wanted to keep that method name available. I sounds like a better fit for 
 getting called by the mapping layer to apply conversion.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoExpression.java
christophstrobl: Two things: 
- static import clashes with eg. `org.bson.Document.parse` 
- the expression must not be evaluated immediately which is something I'd expect from `parse`.

Issue #3590, opened 2021-03-11
Embedded sharding keys are not correctly picked up from the shardKeySource Document


christian-german:
The annotation @Sharded(shardKey = {...}) add some filters to the request performed upon MongoDB when updating a document (upsert).

If we use composite keys like "location.locationId", the value from the Document we store is not retrieved, due to the Document not being "flatted".
This is done in the `applyShardKey` method, in the `UpdateContext` inner class within the `QueryOperations` class:
`getMappedShardKeyFields(domainType).forEach(key -> filterWithShardKey.putIfAbsent(key, shardKeySource.get(key)));`

See [this question](https://stackoverflow.com/questions/66562616/sharding-embedded-document-in-spring-data-mongodb) on SO.

christian-german:
I would be happy to help on this one ;)

christophstrobl:
@christian-german thanks! If you like please have a look at/comment/test #3591.

christian-german:
We have tested our application against the 3.2.0-GH-3590-SNAPSHOT version and everything went fine.

Meanwhile, we have implemented a workaround we will use while waiting for the official release.

From our repository implementation:
```
@Override
	public void saveWithShardKey(DataPayload dataPayload) {

		// If new datapayload
		if(dataPayload.getId() == null) {
			mongoTemplate.insert(dataPayload);

		} else { // If update existing datapayload

			// Criteria according to datapayload shardkey
			Criteria criteria = where("_id").is(new ObjectId(dataPayload.getId()))
					.and("location.locationId").is(dataPayload.getLocation().getLocationId())
					.and("ts").is(dataPayload.getTs());
			Query query = query(criteria);

			// Convert object to document
			Document document = new Document();
			mongoTemplate.getConverter().write(dataPayload, document);
			Update update = Update.fromDocument(document);

			mongoTemplate.upsert(query, update, "dataPayload");
		}
	}
```

PR #3591, opened 2021-03-11
Fix ShardKey lookup for nested paths.

christophstrobl:
This PR fixes the lookup of shard key values for nested paths using the dot (`.`) notation.

Closes: #3590


christian-german:
Looks good to me. Especially the way you check the dotted key in the BsonUtils class to not affect performances for the classic not-nested keys

mp911de:

Issue #3601, opened 2021-03-19
Criteria object not allowing to use field names with underscore in them


SubashL:
Hi,

Our current Spring Data MongoDB version is 2.0.5.RELEASE.

We are migrating to 3.1.5.

Below is the query we use to retrieve the data from the database

@Autowired
private MongoOperations mongoOperation;

query.addCriteria(Criteria.where("business_area").is(buName)
                .andOperator(Criteria.where("onboard_status").exists(true)
                .andOperator(Criteria.where("onboard_status").is("approve"))));

mongoOperation.find(query, ABC.class, "abc");

and these are the imports we use for the above classes.

import org.springframework.data.mongodb.core.MongoOperations;
import org.springframework.data.mongodb.core.query.Criteria;
import org.springframework.data.mongodb.core.query.Query;

In 2.0.5.RELEASE this used to be working fine. We get the desired data. But in 3.1.5. version, we are getting the below error.

**org.springframework.data.mapping.PropertyReferenceException: No property onboard found for type FluidApplication!**

Below is the POJO definition for the above data.

`public class FluidApplication {

    @Id
    @JsonSerialize(using = ToStringSerializer.class)
    private ObjectId id;
    private String requestor_name;
    private String requestor_email;
    private String business_area;
    private String onboard_status;
    private String appType;

    private String application_name;
    private String application_type;
    private String business_functionality;
    private String service_critically;
    private String business_impact;
    private String golivedate;
} `

**Please let us know how to solve this issue. We cannot use @Column annotation above the fields as we have more fields of this kind in other collections and please let us know whether something has changed from 2.0.5.RELEASE to 3.1.5.**

SubashL:
Please assist us in solving the issue. this is blocking our project. We like to know what have changed wrt underscore usage in field names from 2.0.5.RELEASE and 3.1.5

SubashL:
@christophstrobl please let us know if you need any information from us. Could you please let us know what does it mean by type: regression.

christophstrobl:
Can you please provide the full stack trace and context within which the error occurs. Preferable a [Minimal Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example) that demonstrates the issue.

I've been looking current _main_ branch and the mentioned _3.1.5_ release and the following works just fine on both.
```
@Test
void criteriaUsingPropertiesWithUnderscores() {

	WithUnderscoreProperties source = new WithUnderscoreProperties();
[...]

PR #3607, opened 2021-03-25
Fix query mapping resolution of properties using underscore within field name.

christophstrobl:
Closes: #3601

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/QueryMapper.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/QueryMapper.java
@@ -1170,7 +1170,7 @@ public class QueryMapper {
 
 			if (sourceProperty != null && sourceProperty.getOwner().equals(entity)) {
 				return mappingContext
-						.getPersistentPropertyPath(PropertyPath.from(sourceProperty.getName(), entity.getTypeInformation()));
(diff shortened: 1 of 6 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/QueryMapper.java
```diff
@@ -1170,7 +1170,7 @@ public String getMappedKey() {
 
 			if (sourceProperty != null && sourceProperty.getOwner().equals(entity)) {
 				return mappingContext
-						.getPersistentPropertyPath(PropertyPath.from(sourceProperty.getName(), entity.getTypeInformation()));
+						.getPersistentPropertyPath(PropertyPath.from(Pattern.quote(sourceProperty.getName()), entity.getTypeInformation()));
```
mp911de, line 1173: Wouldn't it rather make sense to fix that issue in Commons? Property path creation does all sorts of string transformations and lookups (Decapitalize, `owningType.getProperty(propertyName)`) while regex matching is a small part of the entire process.

christophstrobl: it is yes - for the upcoming service release I'd tough rather have the regression fixed in this project and investigate a fix for commons in the upcoming release cycle as it may lead to unintentional behaviour in other modules.

mp911de:
Follow-up: #3620

mp911de:

PR #3611, opened 2021-03-26
Return saved entity reference instead of original reference.

### LINKED ISSUE #3609: SimpleReactiveMongoRepository#saveAll does not populate @Id property if it is immutable
When using the `Flux<S> saveAll(Publisher<S> entityStream)` method from the `SimpleReactiveMongoRepository` class, the returned entities do not have their `@Id` annotated property populated, in the case where the  annotated field is immutable.

The `@Id` annotated property should obviously be populated with the ObjectId generated by mongodb.

This bug does not affect `Flux<S> saveAll(Iterable<S> entities)` nor `Flux<S> insert(Publisher<S> entities)`. It only occurs if the `@Id` property is immutable (ie: `final` and no setter).

I was able to reproduce the bug on the following versions:
-  spring-data-mongodb `3.1.6`, spring-data-commons `2.4.6`
- spring-data-mongodb `3.0.7`, spring-data-commons `2.3.7`
- spring-data-mongodb `2.2.12`, spring-data-commons `2.2.12`

Here a sample repository to reproduce this bug: [petitcl/spring-data-mongodb-sample-1](https://github.com/petitcl/spring-data-mongodb-sample-1). This sample shows that the bug occurs for an immutable entity but not for a mutable entity. 

After looking at the code of `SimpleReactiveMongoRepository` in version `2.4.4`, it seems to be due to the fact that the original entity reference is returned, instead of the entity returned by the underlying `MongoOperations`.
This works if the entity is mutable, because the entity reference will be updated with the new id, but it does not work if the entity is immutable, as it would require a new reference with the populated id.

Current Code:
```java
	@Override
	public <S extends T> Flux<S> saveAll(Publisher<S> entityStream) {

		Assert.notNull(entityStream, "The given Publisher of entities must not be null!");

		return Flux.from(entityStream)
			.flatMap(
				entity -> entityInformation.isNew(entity) ?
				mongoOperations.insert(entity, entityInformation.getCollectionName()).then(Mono.just(entity)) :
				mongoOperations.save(entity, entityInformation.getCollectionName()).then(Mono.just(entity))
			);
	}
```

What I assume it should look like:
```java
	@Override
	public <S extends T> Flux<S> saveAll(Publisher<S> entityStream) {

		Assert.notNull(entityStream, "The given Publisher of entities must not be null!");

		return Flux.from(entityStream)
			.flatMap(
				entity -> entityInformation.isNew(entity) ?
					mongoOperations.insert(entity, entityInformation.getCollectionName()) :
					mongoOperations.save(entity, entityInformation.getCollectionName())
			);
	}
```

If you validate that this is indeed a bug, I can probably submit a PR to fix it.

petitcl:
Make `SimpleReactiveMongoRepository#saveAll(Publisher<S>)` return the saved entity references instead of the original references.

Closes #3609

This is my first pull request, so I am sorry if am missing anything. Please tell me if something is missing and I will add it gladly.

I added a `ReactiveImmutablePersonRepository` and `ImmutablePerson` classes in `SimpleReactiveMongoRepositoryTests` but it would probably be better to create a dedicated test class that validates that immutable uses cases (including this one) are working.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleReactiveMongoRepository.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleReactiveMongoRepository.java
@@ -50,6 +50,7 @@ import com.mongodb.client.result.DeleteResult;
  * @author Christoph Strobl
  * @author Ruben J Garcia
  * @author Jens Schauder
+ * @author Clément Petit
  * @since 2.0
  */
 public class SimpleReactiveMongoRepository<T, ID extends Serializable> implements ReactiveMongoRepository<T, ID> {
@@ -113,8 +114,8 @@ public class SimpleReactiveMongoRepository<T, ID extends Serializable> implement
 		Assert.notNull(entityStream, "The given Publisher of entities must not be null!");
 
 		return Flux.from(entityStream).flatMap(entity -> entityInformation.isNew(entity) ? //
-				mongoOperations.insert(entity, entityInformation.getCollectionName()).then(Mono.just(entity)) : //
-				mongoOperations.save(entity, entityInformation.getCollectionName()).then(Mono.just(entity)));
+				mongoOperations.insert(entity, entityInformation.getCollectionName()) : //
+				mongoOperations.save(entity, entityInformation.getCollectionName()));
 	}
 
 	/*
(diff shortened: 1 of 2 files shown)


mp911de:
Thanks a lot. We'll take your pull request from here.

mp911de:

PR #3615, opened 2021-03-30
Handle nested Pattern in Criteria equals

### LINKED ISSUE #3414: Criteria or toEquals fail if contains regex [DATAMONGO-2559]
**[Tera](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER49634)** opened **[DATAMONGO-2559](https://jira.spring.io/browse/DATAMONGO-2559?redirect=false)** and commented

Hi

I've found that Criteria object fails equality check if multiple regex criteria are chained into an "orOperator"

 

Here the test i wrote to verify (written in kotlin)

 

```java
class TestCriteriaEquality {
    
    @Test
    fun testSingleRegExEquality() {
        val expected = Criteria("foo").regex("value", "i")
        val actual = Criteria("foo").regex("value", "i")               assertEquals(expected, actual)
    }

    @Test
    fun checkCriteriaOrRegExEquality() {
        val expected = Criteria("a").orOperator(
                Criteria("foo").regex("value", "i"),
                Criteria("bar").regex("value")
        )        val actual = Criteria("a").orOperator(
                Criteria("foo").regex("value", "i"),
                Criteria("bar").regex("value")
        )        Assert.assertEquals(expected, actual)
    }    

    @Test
    fun checkCriteriaOrIsEqualToEquality() {
        val expected = Criteria("a").orOperator(
                Criteria("foo").isEqualTo("value"),
                Criteria("bar").isEqualTo("value")
        )        val actual = Criteria("a").orOperator(
                Criteria("foo").isEqualTo("value"),
                Criteria("bar").isEqualTo("value")
        )        Assert.assertEquals(expected, actual)
    }
}
```

 

 

---

**Affects:** 2.2.7 (Moore SR7)

petitcl:
The equals method of `Pattern` returns false if two `Patterns` are equal but are different instances. This is already handled in the `equals` method of `Criteria` and works well. However, sometimes a `Pattern` can be nested in a `BasicDBList` or in a `Document` or both, like in the case of a `$or` operator.

This PR fixes that: it enhances the `equals` method to manually compare  `BasicDBLists` and `Documents` using the custom `isEqual` method. I am not aware of any util that allow to compare 2 collections using a custom `isEqual` function so I had to implement it manually.

Closes #3414

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Criteria.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Criteria.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Criteria.java
```diff
@@ -989,6 +991,34 @@ private boolean isEqual(Object left, Object right) {
...
 		}
 
+		if (left instanceof Document) {
+			if (!(right instanceof Document)) {
+				return false;
+			}
+			Document leftDocument = (Document) left;
+			Document rightDocument = (Document) right;
+
+			return isEqual(leftDocument.values(), rightDocument.values());
```
mp911de, line 1001: An equality check for `Map`-like types requires a comparison of keys and values. Right now, a document could yield equality if only the order of values is the same regardless of the key the values is assigned to.

petitcl: Sorry, I totally missed that one. I fixed it now: the keys and the values are compared in the case of a `Document`.

mp911de:

Issue #3623, opened 2021-04-11
`@Aggregation` repository query method causes `NullPointerException` when the result is empty


rootdavinalfa:
Hello, 

i'm have an issue regarding usage of Aggregation annotation on spring-data-mongodb

What i'm encounter is 

> Cannot invoke \"org.bson.Document.isEmpty()\" because \"source\" is null

This happen when collection empty (not have any document) but if have an document, i'm not encounter the error.

Here some snippet of code i'm using:

```
@Aggregation(
        "{" +
                "\$project: { maxid: {\$max: '\$_id'} }" +
                "}"
    )
    fun getMaxUnitID(): String?
```

christophstrobl:
Thanks for reporting - seems we miss a `null` check in `AggregationUtils`.

PR #3625, opened 2021-04-12
Fix NPE in declarative aggregation execution.

christophstrobl:
This PR fixes an issue where using a simple return type leads to NPE when the actual aggregation result does not contain any values.

Closes: #3623


mp911de:

PR #3632, opened 2021-04-17
Fix bullet points in aggregations framework asciidoc

petitcl:
There is a small issue in the reference documentation, in the aggregation framework support exactly: some bullet points are not well formatted (see screenshot below). Also there is a small typo (none-xistent instead non-existent).

This PR fixes those 2 smalls issues.

spring-data-mongodb-asciidoc-bullet-list-issue


christophstrobl:
Thank you @petitcl!

Issue #3635, opened 2021-04-23
$floor isOrOrNor() return true 


fuxingjun:
$floor
method isOrOrNor() return true

![企业微信截图_16191698593453](https://user-images.githubusercontent.com/25798786/115850854-e28e8380-a458-11eb-96b7-fcfba68e5bdf.png)
![企业微信截图_16191697364958](https://user-images.githubusercontent.com/25798786/115850869-e8846480-a458-11eb-8611-265bde6779dc.png)

christophstrobl:
Good catch - thanks for reporting!
It would be great if you'd use [code fences](https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks) instead of images next time.

PR #3637, opened 2021-04-26
Fix $or / $nor keyword mapping in query mapper.

christophstrobl:
This PR fixes an issue with the pattern used for detecting `$or` / `$nor` which also matched other keywords like `$floor`.

Closes: #3635


mp911de:

Issue #3633, opened 2021-04-20
NPE in QueryMapper when use Query with `null` as value


EGD:
Spring Data MongoDB version: 3.1.5

used Query annotation:
```java
@Query("parentId: null")
AnyEntity findRootScope();
```
used entity class:
```public class AnyEntity {
    @Id
    public String id;
    public String name;
    @Field(targetType = FieldType.OBJECT_ID)
    public String parentId;
}
```

when calling this method i get NullPointerException:
> java.lang.NullPointerException
	at org.springframework.data.mongodb.core.convert.QueryMapper.applyFieldTargetTypeHintToValue(QueryMapper.java:692)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedValue(QueryMapper.java:335)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedObjectForField(QueryMapper.java:240)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedObject(QueryMapper.java:143)
	at org.springframework.data.mongodb.core.QueryOperations$QueryContext.getMappedQuery(QueryOperations.java:255)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2542)
	at org.springframework.data.mongodb.core.ExecutableFindOperationSupport$ExecutableFindSupport.doFind(ExecutableFindOperationSupport.java:220)
	at org.springframework.data.mongodb.core.ExecutableFindOperationSupport$ExecutableFindSupport.oneValue(ExecutableFindOperationSupport.java:132)
	at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.lambda$getExecution$4(AbstractMongoQuery.java:159)
	at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.doExecute(AbstractMongoQuery.java:132)
	at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.execute(AbstractMongoQuery.java:107)
	at org.springframework.data.repository.core.support.RepositoryMethodInvoker.doInvoke(RepositoryMethodInvoker.java:137)
	at org.springframework.data.repository.core.support.RepositoryMethodInvoker.invoke(RepositoryMethodInvoker.java:121)
	at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.doInvoke(QueryExecutorMethodInterceptor.java:152)
[...]

PR #3643, opened 2021-05-04
Fix NPE in QueryMapper when trying to apply target type on null value.

christophstrobl:
This PR fixes a NPE when a `Query` does a `null` value equality comparison on a field that has a custom target type defined.

```java
@Field(targetType = FieldType.OBJECT_ID) 
String parentId;

// ...

Query q = query(where("parentId").is(null));
template.find(q, ...
```

Further more we introduced additional methods (`isNull()`, `isNullValue()`) on `Criteria` that simplify query creation for `null` or missing fields.

Closes: #3633

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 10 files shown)


mp911de:
That's polished, merged and backported now.

PR #3645, opened 2021-05-06
Issue-3543 - Aggregation annotation should be able to return a Page i…

divyajnu08:
…f you pass Pageable and Sort

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/StringBasedAggregation.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/StringBasedAggregation.java
@@ -20,6 +20,8 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import org.bson.Document;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.SliceImpl;
 import org.springframework.data.mapping.model.SpELExpressionEvaluator;
 import org.springframework.data.mongodb.InvalidMongoDbApiUsageException;
 import org.springframework.data.mongodb.core.MongoOperations;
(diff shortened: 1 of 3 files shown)


divyajnu08:
Hi Mark , 
I have made the changes , now aggregation will return a Slice if pageable is passed.
Kindly check.
Thanks

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/StringBasedAggregation.java
```diff
@@ -118,7 +119,17 @@ protected Object doExecute(MongoQueryMethod method, ResultProcessor resultProces
 
 			return result.getMappedResults();
 		}
-
+		
+		List mappedResults = result.getMappedResults(); 
+		
+		if(method.isPageQuery() || method.isSliceQuery()) {
```
mp911de, line 125: There's a slight difference between `Page` and `Slice`. Paging is aware of the total amount of items while slice requests `pageSize+1` to determine whether there's a next slice available.

Probably paging becomes more difficult because we have to issue a count query and that kind-of imposes performance questions when a complex aggregation is used.

For the time being, let's consider only `Slice` and I'm going to raise the issue with our team to discuss the performance aspect when someone wants to use  `Page<T>`.

divyajnu08: done the changes

mp911de:
Returning a `Page` indeed doesn't make sense as the computation overhead for aggregations outweighs the benefits. The aggregation would require wrapping with a counting stage and would be executed twice. So we will stick with what we have in this pull request.

mp911de:
Thank you for your contribution. That's merged and polished now. I added also support for `Stream`.

PR #3646, opened 2021-05-09
Add an option to @Field annotation to control property write rules

divyajnu08:
…on write

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
@@ -654,10 +654,13 @@ public class MappingMongoConverter extends AbstractMongoConverter implements App
 
 			Object value = accessor.getProperty(prop);
 
-			if (value == null) {
+			if(value==null) {
+				if(!prop.isOmitNullProperty()) {
+					writeSimpleInternal(value, bson , prop);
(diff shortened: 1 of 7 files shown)


divyajnu08:
Hi Christoph/Mark 

Please find the pull request , I hope it solves the issue , kindly let me know if any comments.
Also I could find only one test case , could you please guide me what are the other test cases need to be changed ?

Thanks

mp911de:
Thanks for your pull request. It goes already in the right direction. 

Looking at the code, I'm wondering whether it would actually make sense to represent `null` handling with an enumeration. `omitNull` is a negation of what should happen when writing values and from a wording perspective, I don't have a much more descriptive attribute name that would resolve to `boolean`. 

I suggest expressing what should happen if the value is `null` along the lines of `@Field(write=ALWAYS|NON_NULL)` accompanied with an enum `Field.Write {ALWAYS,NON_NULL}` defaulting to `NON_NULL`. Anything else (omitNull, onNull) could be viewed from both sides, if the property is `null` or the field in the document is `null`/non-existent.

divyajnu08:
Hi Mark 

Kindly find the review comments changes updated on the branch.
Let me know if any other change needs to be done.

Thanks

mp911de:
Thanks a lot. I pushed a polished variant of the changes to https://github.com/spring-projects/spring-data-mongodb/tree/issue/3407

mp911de:

Issue #3602, opened 2021-03-19
Add support for manual references.


christophstrobl:
The goal is support linking documents without the need of using the fixed structure of a `DBRef` (related to: #2780).

### Linking via Entity id
The `@Id` value of the referenced entity is used for linking.
The domain type serves as collection source on read.

```java
class Order {
	
	@Id Long orderId;

	@ManualReference
	Customer customer;
}

class Customer {
	@Id Long customerId;
}
```

```javascript
// write
{
	'_id' : 1,
	'customer' : 100,
}

// read
db.customer.find({ '_id' : 100 }).limit(1)
```

### Linking via custom object

The reference value used for storing is provided via a `DomainType -> ObjectReference` converter.
The `lookup` attribute uses SpEL to create the filter query following the annotated query syntax from `@Query`. The target `collection` that can be either read from the source document, domain type or is a fixed value.

```java
class CustomerReferenceConverter implements Converter<Customer, ObjectReference> {

	public ObjectReference convert(Customer source) {
		return new org.bson.Document("id", source.getId())
			.append("collection", "customers");
	}
}

class Order {
	
	@Id Long orderId;

	@ManualReference(lookup = "{ '_id' : '?#{id}' }", collection = "?#{collection}")
	Customer customer;
}
```

```javascript
// write
{
	'_id' : 1,
	'customer' : {
		'id' : 100,
		'collection' : 'customers'
	}
}

// read
db.customers.find({ '_id' : 100 }).limit(1)
```

The `lookup` filter can contain any valid expression as long as it is resolvable against the stored reference.

### Linking multiple entities

Linking a collection of entities follows the same conversion patterns as described above. The main difference is that the values are stored within an array.
The provided filter expression is created for every entry and concatenated via the `$or` operator.

```java
class Customer {
	
	@Id Long customerId;

	@ManualReference(lookup = "{ '_id' : '?#{target}' }")
	List<Order> orders;
}
```

```javascript
// write
{
	'_id' : 100,
	'orders' : [1,2,3,5,8]
}

// read
db.order.find({ '$or' : [ { '_id' : 1 }, { '_id' : 2 }, ...] })
```

Element order can be messed up when loading documents from the store this way. The original order must be restored by using the filter as a reference without having to fetch objects one by one.

### Lazy Loading

References can be resolved lazily delaying resolution on first access.
In this case on read a proxy is generated holding the actual source value and the lookup filter.

```java
class Order {
	
	@Id Long orderId;

[...]

PR #3647, opened 2021-05-12
Enhance support for linking documents.

christophstrobl:
Using `@DocumentReference` offers an alternative way of linking entities in MongoDB. While the goal is the same as when using `@DBRef`, the store representation is different. Document references, do not follow a specific format. They can be literally anything, a single value, an entire document, basically everything that can be stored in MongoDB. 
By default, the mapping layer will use the referenced entities _id_ value for storage and retrieval, like in the sample below.

```java
@Document
public class Account {

  @Id
  private String id;
  private Float total;
}

@Document
public class Person {

  @Id
  private String id;

  @DocumentReference
  private List<Account> accounts;
}
```

```java
Account account = ...

tempate.insert(account);

template.update(Person.class)
  .matching(where("id").is("6c78727"))
  .apply(new Update().push("accounts").value(account)) 
  .first();
```

```json
{
  "_id" : "6c78727",
  "accounts" : [ "6509b9e", "d3e21be" ] 
}
```

The sample above uses an `_id` based fetch query (`{ '_id' : ?#{#target} }`) for data retrieval and resolves linked entities eagerly.

`@DocumentReference(lookup=...)` allows to define custom queries that are independent from the `_id` field and therefore offer a flexible way of defining links between entities as demonstrated in the sample below, where the `Publisher` of a book is referenced by its acronym instead of the internal `id`.

```java
@Document
public class Book {

  @Id
  private ObjectId id;
  private String title;
  private List<String> author;

  @Field("publisher_ac")
  @DocumentReference(lookup = "{ 'acronym' : ?#{#target} }")
  private Publisher publisher;
}

@Document
public class Publisher {

  @Id
  private ObjectId id;
  private String acronym;
  private String name;

  @DocumentReference(lazy = true)
  private List<Book> books;

}
```

```json
{
  "_id" : "9a48e32",
  "title" : "The Warded Man",
  "author" : ["Peter V. Brett"],
  "publisher_ac" : "DR"
}
```

The above snipped shows the reading side of things when working with custom linked objects. To make the writing part aware of the modified document pointer a custom converter, capable of the transformation into a `DocumentPointer`, like the one below, needs to be registered.

```java
@WritingConverter
class PublisherReferenceConverter implements Converter<Publisher, DocumentPointer<String>> {

	@Override
	public DocumentPointer<String> convert(Publisher source) {
		return () -> source.getAcronym();
	}
}
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.3.0-SNAPSHOT</version>
+	<version>3.3.0-GH-3602-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
--- a/spring-data-mongodb-benchmarks/pom.xml
+++ b/spring-data-mongodb-benchmarks/pom.xml
@@ -7,7 +7,7 @@
 	<parent>
 		<groupId>org.springframework.data</groupId>
 		<artifactId>spring-data-mongodb-parent</artifactId>
-		<version>3.3.0-SNAPSHOT</version>
+		<version>3.3.0-GH-3602-SNAPSHOT</version>
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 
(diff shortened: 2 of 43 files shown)


mp911de:

yihtserns:
https://github.com/spring-projects/spring-data-mongodb/issues/3602

(Manually linking to #3602 because nobody linked them)

Issue #3659, opened 2021-05-26
[3.2.1] Indexing Class with Custom Converter -> Couldn't find PersistentEntity for property private [...]


IsNull:
We updated from spring boot 2.4.2 to 2.5.0, and now Index creation is broken. This means the same code that worked before wont under 2.5.0. (spring-data-mongodb `3.1.3` to `3.2.1` )

----

I have created a sample Project which shows the problem: https://gitlab.com/elderbyte/public/examples/spring-mongo-index-issue

----

Creating an mongo index fails, when the dot path notation is used and the referenced class ~~is in a separate library~~ has a custom mongo converter:

```java
@Configuration
@EnableMongoRepositories()
public class DefaultMongoConfiguration {


    @Bean
    public MongoCustomConversions customConversions(){
        var converters = new ArrayList<Converter<?,?>>();
        converters.add(new MyAddressToDocumentConverter()); // Comment this out, and it works
        return new MongoCustomConversions(converters);
    }

    @WritingConverter
    public static class MyAddressToDocumentConverter implements Converter<MyAddress, Document> {
        @Override
        public Document convert(MyAddress address) {
            var doc = new Document();
            doc.put("street", address.getStreet());
            return doc;
        }
    }
}

```

The entity for which this index is being defined looks like this (simplified):

```java
@Document
public class Customer {

    @Id
    private ObjectId id;

    private String name;

    private MyAddress address;

}
```

```java
public class MyAddress {

    private String street;

    public String getStreet() {
        return street;
    }

    public void setStreet(String street) {
        this.street = street;
    }
}
```

Creating an Index via ensure index yields an unexpected exception:

```java
        indexOps.ensureIndex(
                TextIndexDefinition
                        .builder()
                        .named("TextIndex")
                        .onField("name")
                        .onField("address.street")
                        .build()
        );
```

Stacktrace 
```
org.springframework.data.mapping.MappingException: Couldn't find PersistentEntity for property private com.elderbyte.starter.demo.domain.customers.MyAddress com.elderbyte.starter.demo.domain.customers.Customer.address!
	at org.springframework.data.mapping.context.MappingContext.getRequiredPersistentEntity(MappingContext.java:152) ~[spring-data-commons-2.5.1.jar:2.5.1]
	at org.springframework.data.mapping.context.PersistentPropertyPathFactory.getPair(PersistentPropertyPathFactory.java:225) ~[spring-data-commons-2.5.1.jar:2.5.1]
[...]

PR #3661, opened 2021-06-01
Fix query mapper path resolution for types considered simple ones.

christophstrobl:
spring-projects/spring-data-commons#2293 changed how `PersistentPropertyPath` is resolved and considers potentially registered `Converter` for those, which made the path resolution fail in during the query mapping process.
This PR makes sure to capture the `Exception` and continue with the given user input.

Fixes: #3659


mp911de:

Issue #3225, opened 2019-09-17
Add support for Wildcard Indexes [DATAMONGO-2368]


spring-projects-issues:
**[Christoph Strobl](https://jira.spring.io/secure/ViewProfile.jspa?name=cstrobl)** opened **[DATAMONGO-2368](https://jira.spring.io/browse/DATAMONGO-2368?redirect=false)** and commented

---

**Reference URL:** https://docs.mongodb.com/manual/core/index-wildcard/

PR #3671, opened 2021-06-16
Add support for Wildcard Index

christophstrobl:
A `WildcardIndex` is an index that can be used to include all fields or specific ones based a given (wildcard) pattern.

The index can be set up programmatically using `WildcardIndex` via `IndexOperations`.

```java
mongoOperations
    .indexOps(User.class)
    .ensureIndex(new WildcardIndex("userMetadata"));
```
```javascript
db.user.createIndex({ "userMetadata.$**" : 1 }, {})
```

The `@WildcardIndex` annotation allows a declarative index setup and can be added on either a type or property.

If placed on a type that is a root level domain entity (one having an `@Document` annotation) will advise the index creator to create a wildcard index for it.

```java
@Document
@WildcardIndexed
public class Product {
    ...
}
```
```javascript
db.product.createIndex({ "$**" : 1 },{})
```

The `wildcardProjection` can be used to specify keys to in-/exclude in the index.

```java
@Document
@WildcardIndexed(wildcardProjection = "{ 'userMetadata.age' : 0 }")
public class User {
    private @Id String id;
    private UserMetadata userMetadata;
}
```javascript
db.user.createIndex(
  { "$**" : 1 },
  { "wildcardProjection" :
    { "userMetadata.age" : 0 }
  }
)
```

Wildcard indexes can also be expressed by adding the annotation directly to the field.
Please note that `wildcardProjection` is not allowed on nested paths.

```java
@Document
public class User {
    private @Id String id;

    @WildcardIndexed
    private UserMetadata userMetadata;
}
```
```javascript
db.user.createIndex({ "userMetadata.$**" : 1 }, {})
```

Closes: #3225

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>3.3.0-SNAPSHOT</version>
+	<version>3.3.0-GH-3225-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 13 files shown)


mp911de:

Issue #3668, opened 2021-06-14
Projection on the _id field returns wrong result when using `@MongoId` (MongoDB 4.4)


steschae:
**Describe the bug**
Executing a MongoDB query against MongoDB 4.4 which contains a projection on the _id field with an entity class using the  `@MongoId`  annotation on a string-id field results in returned objects which have incorrect id values. In fact, the id values of all returned objects all have the same value "1".
However, when using the `@Id` instead of the `@MongoId` annotation, the same query returns the expected results containing the correct id values.

**Expected behavior**
The returned objects each have their correct id value.

**Actual behavior**
When using MongoDB 4.4: The returned objects each have the id value "1"
When using MongoDB 4.2: The returned objects each have their correct id value.

**To Reproduce**
I have set up an example project with test cases (https://github.com/steschae/spring-data-mongodb-id-projection-bug). The test "MongoDbVersion4_4_Test > should_find_all_person_ids_when_using_mongo_id_annotation" is the one which is failing because of this bug (https://github.com/steschae/spring-data-mongodb-id-projection-bug/runs/2818935410#step:4:33).

To reproduce, 1. create an entity class using the `@MongoId` annotation
```java
@Data
public class Person {
    @MongoId
    private String id;
}
```

2. Create a query with a projection on the _id field:
```java
public interface PersonRepository extends CrudRepository<PersonWithMongoId, String> {
    @Query(value="{}", fields = "{ _id : 1 }")
    List<Person> findAllIds();
}
```
or
```java
final var query = new Query();
query.fields().include(ID);
mongoTemplate.find(query, Person.class);
```

3. When executing the query, all returned objects have the value "1" in the id field.

![image](https://user-images.githubusercontent.com/79054136/121884941-d4037080-cd13-11eb-814b-2717d9f3bc75.png)

**Configuration/Environment/Context**
MongoDB 4.4
org.springframework.data:spring-data-mongodb:3.2.1
org.mongodb:mongodb-driver-sync:4.2.3

**Additional information**
With MongoDB 4.4 the behaviour of projection documents in MongoDB queries has changed:
(https://docs.mongodb.com/manual/release-notes/4.4-compatibility/#projection-compatibility-changes)

The query is translated to `db.getCollection("persons").find({}, {"_id": "1"})` (the value of the _id projection is a string). This worked with MongoDB 4.2, but with MongoDB 4.4 the string value "1" in the projection causes the id value of the returned documents to be overidden with this value "1".
[...]

PR #3678, opened 2021-06-21
Fix field projection value conversion.

christophstrobl:
The field projection conversion should actually only map field names and avoid value conversion. In the `@MongoId` case a numeric inclusion parameter `1` was unintentionally converted into its `String` representation `"1"` which causes trouble on Mongo 4.4 servers.

Fixes: #3668


mp911de:

PR #3687, opened 2021-06-25
Fix build on Java 16

christophstrobl:
Fix mapping and context setup in tests.
Use spring-projects/spring-data-commons#2390 to get rid of optional mapping.
Disable tests that require class proxies on core java types like `ArrayList`.

depends-on: spring-projects/spring-data-commons#2390
Resolves: #3656


mp911de:
That's merged now.

Issue #3686, opened 2021-06-25
reading a document with a list with a null element fails with Spring Data Mongo 3.2.2, works with 3.2.1


jnaalisv:
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "source" is null

	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$ConversionContext.convert(MappingMongoConverter.java:1885)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readCollectionOrArray(MappingMongoConverter.java:1139)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$ConversionContext.convert(MappingMongoConverter.java:1900)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$MongoDbPropertyValueProvider.getPropertyValue(MappingMongoConverter.java:1636)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readProperties(MappingMongoConverter.java:477)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.populateProperties(MappingMongoConverter.java:392)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.read(MappingMongoConverter.java:371)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readDocument(MappingMongoConverter.java:341)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$ConversionContext.convert(MappingMongoConverter.java:1918)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$MongoDbPropertyValueProvider.getPropertyValue(MappingMongoConverter.java:1636)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readProperties(MappingMongoConverter.java:477)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.populateProperties(MappingMongoConverter.java:392)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.read(MappingMongoConverter.java:371)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readDocument(MappingMongoConverter.java:341)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$ConversionContext.convert(MappingMongoConverter.java:1918)
[...]

PR #3691, opened 2021-07-01
Fix NPE when reading/mapping null value inside collection.

christophstrobl:
Closes: #3686

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
@@ -2006,12 +2006,16 @@ public class MappingMongoConverter extends AbstractMongoConverter implements App
 		/**
 		 * Converts a source object into {@link TypeInformation target}.
 		 *
-		 * @param source must not be {@literal null}.
+		 * @param source can be {@literal null}.
 		 * @param typeHint must not be {@literal null}.
 		 * @return the converted object.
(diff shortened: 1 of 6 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
```diff
@@ -2006,12 +2006,16 @@ public boolean isAssignableFrom(org.springframework.data.util.TypeInformation ta
...
-		 * @param source must not be {@literal null}.
+		 * @param source can be {@literal null}.
 		 * @param typeHint must not be {@literal null}.
 		 * @return the converted object.
 		 */
 		@SuppressWarnings("unchecked")
-		public <S extends Object> S convert(Object source, TypeInformation<? extends S> typeHint) {
+		public <S extends Object> S convert(@Nullable Object source, TypeInformation<? extends S> typeHint) {
+
+			if(source == null) {
```
mp911de, line 2016: Shouldn't that be rather caught in the map and collection iterations as we basically make `convert` nullable?

christophstrobl: The `MappingMongoConverter` does not follow the `o.s.c.c.converter.Converter` contract and accepted `null` in previous versions, though the documentation did state it differently. Since it's a public API, I'd rather restore the original behaviour and alter the documentation, than spread `null` checks in various places.

Issue #3702, opened 2021-07-11
`MappingMongoConverter` incorrectly processes an object property of type `org.bson.Document`


meloujir:
Hi, we encountered a problem (originates from 3.1.5 > 3.2 update), which occurrs when a property (of some DBO entity) of type org.bson.Document is handled within MappingMongoConverter.ConversionContext#convert. MongoDB driver correctly returns a document, of which all inner object properties (fields) are of type org.bson.Document.  

Then MappingMongoConverter comes into play (within process of populating fields , #populateProperties -> #readProperties). During that process, the property value of type org.bson.Document is incorrectly handled as a Map and all properties (fields) with an object value are converted to a LinkedHashMap. Later on, when we, within concrete context, try to convert that object of type org.bson.Document to certain class (using MongoTemplate#getConverter()#read(getEntityClass(), sourceObject), but it is not important), it leads to:

`org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [java.util.LinkedHashMap<?, ?>] to type [...]`

That error is only consequence of the way how the object of property of type org.bson.Document is treated. Yes, org.bson.Document is implementing a Map, but it should not be processed as a Map. 

If we compare codes of version 3.1.5 and version 3.2.X

First, let's look at version 3.1.5:

MappingMongoConverter#readValue

```
@Nullable
@SuppressWarnings("unchecked")
<T> T readValue(Object value, TypeInformation<?> type, ObjectPath path) {

	Class<?> rawType = type.getType();

	if (conversions.hasCustomReadTarget(value.getClass(), rawType)) {
		return (T) conversionService.convert(value, rawType);
	} else if (value instanceof DBRef) {
		return potentiallyReadOrResolveDbRef((DBRef) value, type, path, rawType);
	} else if (value instanceof List) {
		return (T) readCollectionOrArray(type, (List<Object>) value, path);
	} else if (value instanceof Document) {
		return (T) read(type, (Document) value, path); // this "if branch" comes into play if a property of type org.bson.Document is read.
	} else if (value instanceof DBObject) {
		return (T) read(type, (BasicDBObject) value, path);
	} else {
		return (T) getPotentiallyConvertedSimpleRead(value, rawType);
	}
}
```

Then MappingMongoConverter#read is called and the "if (Document.class.isAssignableFrom(rawType))" is chosen as a processing branch).

```
@Nullable
@SuppressWarnings("unchecked")
[...]

Issue #3692, opened 2021-07-01
ClassCastException in MappingMongoConverter when having Map<String, Object> as a field type.


Taha-Di-Nero:
After upgrading to Spring Boot from 2.4.4 to 2.5.2(spring-data-mongodb 3.1.6 --> 3.2.2) I am having the following exception:
```bash
Caused by: java.lang.ClassCastException: java.util.LinkedHashMap incompatible with org.bson.conversions.Bson
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$ConversionContext.convert(MappingMongoConverter.java:1905)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter$MongoDbPropertyValueProvider.getPropertyValue(MappingMongoConverter.java:1636)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readProperties(MappingMongoConverter.java:477)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.populateProperties(MappingMongoConverter.java:392)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.read(MappingMongoConverter.java:371)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readDocument(MappingMongoConverter.java:341)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.read(MappingMongoConverter.java:277)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.read(MappingMongoConverter.java:273)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.read(MappingMongoConverter.java:102)
	at org.springframework.data.mongodb.core.MongoTemplate$ReadDocumentCallback.doWith(MongoTemplate.java:3178)
	at org.springframework.data.mongodb.core.MongoTemplate.executeFindMultiInternal(MongoTemplate.java:2813)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2543)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2525)
	at org.springframework.data.mongodb.core.MongoTemplate.find(MongoTemplate.java:847)
	at org.springframework.data.mongodb.repository.support.SimpleMongoRepository.findAll(SimpleMongoRepository.java:428)
	at org.springframework.data.mongodb.repository.support.SimpleMongoRepository.findAllById(SimpleMongoRepository.java:162)
[...]

PR #3704, opened 2021-07-12
Fix raw document conversion in Collection like properties.

christophstrobl:
Closes: #3702, #3692


mp911de:

PR #3732, opened 2021-07-15
Add support for creating Time Series collection.

christophstrobl:
Introduce time series to `CollectionOptions` and add dedicated `@TimeSeries` annotation to derive values from.

```java
@TimeSeries(collection="weather", timeField = "timestamp")
public class Measurement {
    String id;
    Instant timestamp;
    // ...
}

template.createCollection(Measurement.class);
```

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.mongodb.core.mapping;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.core.annotation.AliasFor;
+import org.springframework.data.mongodb.core.timeseries.Granularities;
+
+/**
+ * Identifies a domain object to be persisted to a MongoDB Time Series collection.
+ *
+ * @author Christoph Strobl
+ * @since 3.3
+ * @see <a href=
+ *      "https://docs.mongodb.com/manual/core/timeseries-collections">https://docs.mongodb.com/manual/core/timeseries-collections</a>
+ */
+@Inherited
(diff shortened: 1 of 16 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
```diff
@@ -0,0 +1,86 @@
...
+	@AliasFor(annotation = Document.class, attribute = "collection")
+	String collection() default "";
+
+	/**
+	 * The name of the property which contains the date in each time series document. <br />
+	 * {@link Field#name() Annotated fieldnames} will be considered during the mapping process.
+	 *
+	 * @return never {@literal null}.
+	 */
+	String timeField();
```
mp911de, line 58: time field vs. time property?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * The name of the field which contains metadata in each time series document. Should not be the {@literal id} nor
+	 * {@link #timeField()} nor point to an {@literal array} or {@link java.util.Collection}. <br />
+	 * {@link Field#name() Annotated fieldnames} will be considered during the mapping process.
+	 *
+	 * @return empty {@link String} by default.
+	 */
+	String metaField() default "";
```
mp911de, line 67: how about just meta or meta property?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
```diff
@@ -0,0 +1,86 @@
...
+ * @author Christoph Strobl
+ * @since 3.3
+ * @see <a href=
+ *      "https://docs.mongodb.com/manual/core/timeseries-collections">https://docs.mongodb.com/manual/core/timeseries-collections</a>
+ */
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+@Document
+public @interface TimeSeries {
```
mp911de, line 39: Alternatively, we could have annotations for the timeseries field and the meta field, similar to text-indexed language for consistency.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
christophstrobl: Staying close to the MongoDB terminology for Time Series collections seemed to be a valid choice here.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
christophstrobl: Staying close to the MongoDB terminology for Time Series collections seemed to be a valid choice here. Simply `meta` does IMO not express that this is a field reference rather than a proved meta object/value.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/TimeSeries.java
christophstrobl: The `@Language` field allows to override the defaults defined via `@Document` so I do not think this is a good fit here since you may not override any of those fields on a per document basis for Time Series collections where it's rather a fixed value at collection creation time.

mp911de:

Issue #3715, opened 2021-07-14
Add support for $rank & $denseRank aggregation operators.


christophstrobl:
Available since MongoDB 5.0.

https://docs.mongodb.com/manual/reference/operator/aggregation/denseRank/

----

Depends on: #3711

Issue #3727, opened 2021-07-14
Add support for $shift aggregation operator.


christophstrobl:
Available since MongoDB 5.0

* https://docs.mongodb.com/manual/reference/operator/aggregation/shift/

----

Depends on #3711

Issue #3717, opened 2021-07-14
Add support for $documentNumber aggregation operator.


christophstrobl:
Available since MongoDB 5.0

https://docs.mongodb.com/manual/reference/operator/aggregation/documentNumber/

----

Depends on: #3711

PR #3741, opened 2021-07-22
Add support for $rank, $denseRank, $documentNumber & $shift aggregation operators.

christophstrobl:
Closes: #3715, #3717, #3727


mp911de:

Issue #3713, opened 2021-07-14
Add support for `$dateAdd` and `$dateDiff` aggregation operators


christophstrobl:
Available since MongoDB 5.0

* https://docs.mongodb.com/manual/reference/operator/aggregation/dateAdd/
* https://docs.mongodb.com/manual/reference/operator/aggregation/dateDiff/

PR #3748, opened 2021-07-26
Add support for `$dateAdd` and `$dateDiff` aggregation operators

christophstrobl:
Closes: #3713


mp911de:

PR #3761, opened 2021-08-03
Add Criteria infix functions for `maxDistance` and `minDistance`

sangyongchoi:
add Criteria infix func 
- maxDistance
- minDistance

Because to use minDistance and maxDistance, near and nearSphere must exist.
But currently using infix fun to use minDistance, maxDistance throws an error.
Because we're adding it to the new Criteria.

[A repository where you can check for errors](https://github.com/sangyongchoi/mongodb-error)

I think good if criteria with Near Sphere and NearSphere added min Distance and Max Distance could be easily added. so I will send you a PR.


christophstrobl:
Thanks @sangyongchoi for  taking the time to report this and craft a fix.

mp911de:

PR #3765, opened 2021-08-07
GH-3726: Add support for $sampleRate match expression.

### LINKED ISSUE #3726: Add support for $sampleRate aggregation operator.
Available since MongoDB 4.4.2.

* https://docs.mongodb.com/manual/reference/operator/aggregation/sampleRate/

JamesMcNee:
Closes #3726

Hey 👋, hope this implementation is good, this new `$sampleRate` expression threw me a little as it looks like the first one we have that can only be applied within the match pipeline stage and is not available outside of aggregation (so not a `Criteria`).


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/Aggregation.java
```diff
@@ -499,6 +500,17 @@ public static MatchOperation match(CriteriaDefinition criteria) {
...
 	}
 
+	/**
+	 * Creates a new {@link MatchOperation} using the given {@link MatchExpression}.
+	 *
+	 * @param matchExpression must not be {@literal null}.
+	 * @return new instance of {@link MatchOperation}.
+	 * @since 3.2.4
+	 */
+	public static MatchOperation match(MatchExpression matchExpression) {
```
mp911de, line 510: We should not introduce another abstraction.

mp911de (changes requested):
I'm reluctant introducing another abstraction for `$sampleRate`. We should find a better representation. Right now, `$sampleRate` is applicable to aggregations only, while the mentioned `$rand` operators works in normal queries. I anticipate that sooner or later, `$sampleRate` would be officially allowed as criteria operator. MongoDB 4.4. accepts already `$sampleRate` via `db.getCollection(…).find({$sampleRate:0.33})`.

Therefore, I'd ask you to update this pull request to introduce a `sampleRate` operator to the `Criteria` type. Doing so would significantly reduce complexity of this change.

PR #3777, opened 2021-08-21
Relax requirement for `GeoJsonMultiPoint` construction allowing creation using a single point

### LINKED ISSUE #3776: Relax requirement for `GeoJsonMultiPoint` construction allowing creation using a single point
I stumbled upon a limitation in GeoJsonMultipoint which I find unnecessary - it requires to have at least 2 points. While there are no requirements in GeoJson RFC for MultiPoint to have this and Mongo itself works just fine with 1 point in the coordinates list

Was there a reason to implement it like this? Will it cause backward compatibility issues if this requirement is relaxed?

tinedel:
Only 1 point is required per GeoJson RFC and Mongo works just fine with 1 point as well.

Closes #3776


mp911de:

Issue #3779, opened 2021-08-23
Factory beans should not contain protected setters.


christophstrobl:
Setters in factory beans like `MongoRepositoryFactoryBean` should be `public`.

PR #3780, opened 2021-08-23
Change visibility of Reactive/MongoRepositoryFactoryBean setters.

christophstrobl:
Setters of the FactoryBean should be public.

Closes: #3779


mp911de:
That's merged and forward-ported now.

Issue #3805, opened 2021-09-08
@DocumentReference empty list problem


Izimbergenov:
Hello,

I have a problem when using @DocumentReference with an empty list. Specifically when we initialize our list as new ArrayList<>().  Below are the test codes: the first one failed, the second one success

	static class Book
	{
		@MongoId(targetType = FieldType.OBJECT_ID)
		String id;

		String name;
	}

	static class Library
	{
		@MongoId(targetType = FieldType.OBJECT_ID)
		String id;

		@DocumentReference
		private List<Book> books = new ArrayList<>();
	}

Failed test code(throws NoSuchElementException when template.findById):

	@Test
	public void Given_Library_Without_Books_When_Get_By_Id_Then_Success() throws Exception
	{
		Library library = new Library();
		library = template.save(library);

		Library existingLibrary = template.findById(library.id, Library.class);
		assertThat(existingLibrary).isNotNull();
	}

Success test code:

	@Test
	public void Given_Library_With_Books_When_Get_By_Id_Then_Success() throws Exception
	{
		Book book = new Book();
		book.name = "Harry Potter";
		book = template.save(book);

		Library library = new Library();
		library.books = Arrays.asList(book);
		library = template.save(library);

		Library existingLibrary = template.findById(library.id, Library.class);
		assertThat(existingLibrary).isNotNull();
	}

Spring Data MongoDB version: 3.3.0-SNAPSHOT

christophstrobl:
Thanks for reporting - we'll look into it.

PR #3807, opened 2021-09-08
Fix document reference on empty reference arrays. 

christophstrobl:
Closes: #3805


mp911de:

PR #3812, opened 2021-09-08
Fix slice argument in query fields projection.

christophstrobl:
Fixes: #3811

----

Should be back ported to maintenance branches.


mp911de:

Issue #3871, opened 2021-11-03
Backwards incompatibility in 2.5.6


sjerman:
Hi Folks,

Trying to bring our application forward to 2.5.6 and am running into an issue with @Query.

Had same issue last time I did the migration - #3401 .. maybe the fix didn't make it to 2.5

The query was:
```
  @Query(
      " ?#{ hasRole('ROLE_ADMIN') or global.getProperty('user.global') == 'true' ? {'_class': { '$eq' : 'com.kloudspot.jamesons.datamodel.domain.location.Region' }} : { '$and' : { {'_class': { '$eq' : 'com.kloudspot.jamesons.datamodel.domain.location.Region' }}, {'$or': { { 'users._id': global.getOid( principal.id ) },{'usersFromHierarchy._id': global.getOid( principal.id ) } } } } } } ")
  @Cacheable(value = "siteRegions", keyGenerator = "siteKeyGenerator")
  List<Region> findRegions();
```
Get this error:

```
org.bson.BsonInvalidOperationException: readStartDocument can only be called when CurrentBSONType is DOCUMENT, not when CurrentBSONType is UNDEFINED.
	at org.bson.AbstractBsonReader.verifyBSONType(AbstractBsonReader.java:690)
	at org.bson.AbstractBsonReader.checkPreconditions(AbstractBsonReader.java:722)
	at org.bson.AbstractBsonReader.readStartDocument(AbstractBsonReader.java:450)
	at org.springframework.data.mongodb.util.json.ParameterBindingDocumentCodec.decode(ParameterBindingDocumentCodec.java:223)
	at org.springframework.data.mongodb.util.json.ParameterBindingDocumentCodec.captureExpressionDependencies(ParameterBindingDocumentCodec.java:202)
	at org.springframework.data.mongodb.repository.query.StringBasedMongoQuery.getBindingContext(StringBasedMongoQuery.java:136)
	at org.springframework.data.mongodb.repository.query.StringBasedMongoQuery.createQuery(StringBasedMongoQuery.java:121)
	at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.doExecute(AbstractMongoQuery.java:122)
	at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.execute(AbstractMongoQuery.java:107)
	at org.springframework.data.repository.core.support.RepositoryMethodInvoker.doInvoke(RepositoryMethodInvoker.java:137)
[...]

PR #3874, opened 2021-11-11
Fix parse error when collecting SpEL extensions for queries computed entirely via expression.

christophstrobl:
This PR makes sure the extension collector returns a marker object to indicate no further json parsing is required.

Fixes: #3871


christophstrobl:
superseded by #3907

PR #3887, opened 2021-11-17
Convert nested elements of a raw Document on write.

christophstrobl:
*** Draft ***


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
```diff
@@ -1002,6 +1002,12 @@ protected Bson writeMapInternal(Map<Object, Object> obj, Bson bson, TypeInformat
 			if (conversions.isSimpleType(key.getClass())) {
 
 				String simpleKey = prepareMapKey(key);
+				if(val instanceof Document) {
+					Document document = new Document();
+					writeMapInternal((Map<Object, Object>) val, document, ClassTypeInformation.MAP);
+					BsonUtils.addToMap(bson, simpleKey, document);
+					continue;
```
TheHett, line 1009: Perhaps it is better to write everything in the same style: either `continue` , or `if .. else`.

christophstrobl:
we closed the related ticket and won't change the current behaviour.

PR #3892, opened 2021-11-29
Fix pagination using the reactive fluent Querydsl API

rolag-it:
Pageable object was not passed to Query, so fetchPage retrieved erroneously the whole dataset as Page content.


mp911de:

PR #3904, opened 2021-12-07
Fix field inclusion in aggregation project operation.

christophstrobl:


mp911de:

PR #3912, opened 2021-12-10
Handling collections and map types before running an element converter

nexx512:
When the custom conversions are checked before collection and map converters, it is not possible to use custom converters for lists and maps.
If a custom converter for a list or a map is defined, it is executed here and not from within the `collectionConverter` where it should be executed.
So I think collections and maps should be handled beforehand.
Maybe lines 2064-2066 might not be needed ad all as this case is covered with line 2081.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java
@@ -2032,10 +2032,6 @@ public class MappingMongoConverter extends AbstractMongoConverter implements App
 			Assert.notNull(source, "Source must not be null");
 			Assert.notNull(typeHint, "TypeInformation must not be null");
 
-			if (conversions.hasCustomReadTarget(source.getClass(), typeHint.getType())) {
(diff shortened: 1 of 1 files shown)


nexx512:
I extendend this functionality to cover lists and maps of custom types, even allowing for collection converters whose target is not a collection. Being not so restrictive leaved more options to the user to implement custom converters.

It should now be possible to map vavr lists and maps with a simple converter like
```java
@ReadingConverter
public ListReadingConverter implements Converter<List<ListType>, io.vavr.collection.List<ListType>> {
        @Override
        public io.vavr.collection.List<ListType> convert(List<ListType> source) {
            return io.vavr.collection.List.ofAll(source);
        }
}
```

However for maps to work correctly, the change from https://github.com/spring-projects/spring-data-commons/pull/2517 is required.

I also removed some mapping tests that tested a mapping with a class extending `java.util.Map`. The extending class however didn't expose a map like interface. So i would say that this case should be covered by a separate class containing a map instead of extending one. This pretty much looked like a good example where aggregation should be preferred over inheritance.

Would be great if someone can have a look at it and discuss it if there are concerns.

odrotbohm:
Relates to spring-projects/spring-data-commons#2511.

Issue #3914, opened 2021-12-16
MongoPersistentEntityIndexResolver should not traverse Map structures.


christophstrobl:
Setting up indices on `Map` like structures requires the _map key_, which cannot be resolved based on the plain type information to be part of the index. Therefore, traversal should not follow the path of the _map value_ type to set up indices that might not be effective within the store.

Relates to: #3869

Issue #3901, opened 2021-12-06
`NullPointerException` in `MongoPersistentEntityIndexResolver.appendTextIndexInformation(…)`


caspianb:
Since upgrading to 2.6.1 I am seeing failures when invoking IndexResolver::resolveIndexFor(...) for some specific entities.

We have a generic pattern using javax.measure classes using an enum which implements an interface defined in this way:

```
public interface Uom<T extends javax.measure.Quantity.Quantity<T>> {
   javax.measure.Unit.Unit<T> getUnit();
}

public enum UomLength implements Uom<javax.measure.quantity.Length>{
   ...
}

// This is the embedded object type we actually persist.
public class UomMeasure<T extends Uom...> {
   T unit; // the enums are stored here
}
```

Pre 2.6, index resolution worked just fine on entities containing this object. However, after upgrading to 2.6.1 I get a NullPointerException with the following stack trace (Note: I can "fix" the issue by marking getUnit() as @Transient which is an acceptable work-around for our use case -- the persistence mechanism still persists the enum via its string properly and we do not use the annotation based index resolution pattern most of the time).

When I debug into the code and view the state at the time of the NPE I see this:
```
entity = mappingContext.getPersistentEntity(persistentProperty.getActualType());
persistentProperty.getActualType() -> Integer.class
entity == null
```

```
java.lang.NullPointerException
	at org.springframework.data.mongodb.core.index.MongoPersistentEntityIndexResolver.appendTextIndexInformation(MongoPersistentEntityIndexResolver.java:338)
	at org.springframework.data.mongodb.core.index.MongoPersistentEntityIndexResolver.access$100(MongoPersistentEntityIndexResolver.java:78)
	at org.springframework.data.mongodb.core.index.MongoPersistentEntityIndexResolver$1.doWithPersistentProperty(MongoPersistentEntityIndexResolver.java:370)
	at org.springframework.data.mongodb.core.index.MongoPersistentEntityIndexResolver$1.doWithPersistentProperty(MongoPersistentEntityIndexResolver.java:338)
	at org.springframework.data.mapping.model.BasicPersistentEntity.doWithProperties(BasicPersistentEntity.java:360)
	at org.springframework.data.mongodb.core.index.MongoPersistentEntityIndexResolver.appendTextIndexInformation(MongoPersistentEntityIndexResolver.java:338)
[...]

PR #3915, opened 2021-12-17
Avoid creating invalid index definitions for `Map`-like properties

christophstrobl:
This commit makes sure to exclude `Map` like structures from index inspection unless annotated with `@WilcardIndexed`.

Resolves: #3914, #3901


mp911de:

Issue #3918, opened 2021-12-23
DocumentReference invokes `find` query twice


TheHett:
Simple `DocumentReference` invokes two queries to mongodb.
It doesn't matter if it's lazy or not.

```
    @DocumentReference(lazy = true)
    @Field("container_id")
    lateinit var container: Container
```

```
2021-12-23 12:15:35.123 DEBUG 1 --- [atcher-worker-1] org.mongodb.driver.protocol.command      : Sending command '{"find": "container", "filter": {"_id": {"$oid": "61b482c8864a10008a452f83"}}, "limit": 1, "singleBatch": true, "$db": "fs", "lsid": {"id": {"$binary": {"base64": "GGCSBmQUTRK7GCofLZuD/w==", "subType": "04"}}}}' with request id 12 to database fs on connection [connectionId{localValue:3, serverValue:15314}] to server fs-mongodb:27017
2021-12-23 12:15:35.124 DEBUG 1 --- [atcher-worker-1] org.mongodb.driver.protocol.command      : Execution of command with request id 12 completed successfully in 1.10 ms on connection [connectionId{localValue:3, serverValue:15314}] to server fs-mongodb:27017
2021-12-23 12:15:35.125 DEBUG 1 --- [atcher-worker-1] org.mongodb.driver.protocol.command      : Sending command '{"find": "container", "filter": {"_id": {"$oid": "61b482c8864a10008a452f83"}}, "limit": 1, "singleBatch": true, "$db": "fs", "lsid": {"id": {"$binary": {"base64": "GGCSBmQUTRK7GCofLZuD/w==", "subType": "04"}}}}' with request id 13 to database fs on connection [connectionId{localValue:3, serverValue:15314}] to server fs-mongodb:27017
2021-12-23 12:15:35.126 DEBUG 1 --- [atcher-worker-1] org.mongodb.driver.protocol.command      : Execution of command with request id 13 completed successfully in 0.81 ms on connection [connectionId{localValue:3, serverValue:15314}] to server fs-mongodb:27017
```

TheHett:
Can you explain, what is QPerson and other prefixed with `Q` files?
Is it should be generated?  IDEA can't compile tests, I wanted to start debug.

![image](https://user-images.githubusercontent.com/946246/147345203-c3b89276-0f76-4a3f-a197-2fa5c1aa7d1c.png)

PR #3919, opened 2021-12-23
Issue 3918: fixed double call fetch method in DefaultReferenceResolver

TheHett:
Issue https://github.com/spring-projects/spring-data-mongodb/issues/3918
No more tests cases provided, because I suspicious that is unnecessary.


christophstrobl:
Good catch! Thank you!

christophstrobl:
Good catch! Thank you!

christophstrobl:
Merged to _main_ and back ported to  _3.3.x_.

PR #3930, opened 2022-01-11
Use index instead of iterator to map positional and map keys for updates.

christophstrobl:


mp911de:

PR #3931, opened 2022-01-12
Avoid schema keyId uuid representation errors. 

christophstrobl:
To avoid driver configuration specific `UUID` representation format errors (binary subtype 3 vs. subtype 4) we now directly convert the given key into its subtype 4 format.


christophstrobl:
don't merge yet - needs a bit more refinement.

PR #3942, opened 2022-01-19
Observability with Micrometer & Micrometer Tracing

marcingrzejszczak:


mp911de:
We should add an integration test (is there a mock `Tracer` implementation) and a bit of documentation regarding the tracing integration.

gregturn:
Where does `MongoTracingRecordingHandler` get wired into everything? I don't see an instance created anywhere but inside test cases.

marcingrzejszczak:
Yes, it will be injected into micrometer#config#timerRecordingHandlers

marcingrzejszczak:
I've updated the code to use the enum based approach to describe metrics and spans. I've added automated docs generation from those sources the outcome of which would be the following doc

image

gregturn:
Resolved via 4c77763cd39abf2a2d7e199778016728b45d79fc

PR #3946, opened 2022-01-24
Use Java 8`Stream` as return type for Template operations returning a stream

### LINKED ISSUE #3944: Use Java 8 `Stream` in `stream` and `aggregateStream` Template methods
With the usage of Java 17 as baseline we want to switch to Java 8 streams directly to avoid the indirection through `CloseableIterator`. We also need to document that the stream needs to be closed (e.g. to be used with try-with-resources).

mp911de:
We now use Stream instead of `CloseableIterator` for easier stream creation.

Closes #3944


christophstrobl:
LGTM

christophstrobl:
Merged to _4.0.x_

Issue #3916, opened 2021-12-17
Update list of supported query method keywords.


christophstrobl:
`IgnoreCase` is not listed in section `mongodb.repositories.queries` for example.

mihailcornescu:
Hi @christophstrobl. I am new to contributing to spring-data-mongodb. Can I handle this issue?

mihailcornescu:
Thank you @christophstrobl. I have checked them already. Can you tell me for which milestone should I include the changes? In other words or from which branch should I create my feature branch.

christophstrobl:
Please use _main_. We'll take care of porting the changes.

mihailcornescu:
@christophstrobl I have raised PR #3950. Can you have a look on it?

PR #3950, opened 2022-01-25
Update list of supported query method keywords.

mihailcornescu:
Fix for #3916.
- included `@IgnoreCase` to mongodb.repositories.queries query table documentation.

### DIFF
--- a/src/main/asciidoc/reference/mongo-repositories.adoc
+++ b/src/main/asciidoc/reference/mongo-repositories.adoc
@@ -283,8 +283,8 @@ lower / upper bounds (`$gt` / `$gte` & `$lt` / `$lte`) according to `Range`
 | `{"location" : {"$exists" : exists }}`
 
 | `IgnoreCase`
-| `findByFirstnameIgnoreCase(String firstname)`
-| `{"firstname" : {"$strcasecmp" : firstname }}`
+| `findByUsernameIgnoreCase(String username)`
+| `{"username" : {"$strcasecmp" : username }}`
 |===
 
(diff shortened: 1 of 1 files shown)


src/main/asciidoc/reference/mongo-repositories.adoc
```diff
@@ -281,6 +281,10 @@ lower / upper bounds (`$gt` / `$gte` & `$lt` / `$lte`) according to `Range`
 | `Exists`
 | `findByLocationExists(boolean exists)`
 | `{"location" : {"$exists" : exists }}`
+
+| `IgnoreCase`
+| `findByUsernameIgnoreCase(String username)`
+| `{"username" : {"$strcasecmp" : username }}`
```
christophstrobl, line 287: `$strcasecmp` is an aggregation operator. The `MongoQueryCreator` uses a `Pattern` that translates to `{ $regex : '^username$', '$options' : 'i' }`.

mihailcornescu: Thank you for the feedback. I will make the changes.

mihailcornescu: Hi @christophstrobl. I have made the changes. You were right, that regex is used in the code.
Is it ok to use double quotes instead of single quotes? I noticed that strings are used and even the i from the options is a String.

christophstrobl:
Thanks for the contribution @mihailcornescu! Merged to _main_ development line and back ported to _3.3.x_ and _3.2.x_.

PR #3951, opened 2022-01-27
Allow to estimate document count. 

christophstrobl:
Introduce an option that allows users to opt in on using `estimatedDocumentCount` instead of `countDocuments` in case the used filter query is empty.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
@@ -22,13 +22,13 @@ import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.util.*;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiPredicate;
 import java.util.stream.Collectors;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.bson.Document;
 import org.bson.conversions.Bson;
-
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
@@ -188,6 +188,8 @@ public class MongoTemplate implements MongoOperations, ApplicationContextAware, 
 
 	private SessionSynchronization sessionSynchronization = SessionSynchronization.ON_ACTUAL_TRANSACTION;
 
+	private CountExecution countExecution = this::doPreciseCount;
+
 	/**
 	 * Constructor used for a basic template configuration.
 	 *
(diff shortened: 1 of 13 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ReactiveMongoOperations.java
```diff
@@ -951,8 +951,11 @@ default ReactiveSessionScoped withSession(Supplier<ClientSession> sessionProvide
 	 *          {@literal null}.
 	 * @param entityClass class that determines the collection to use. Must not be {@literal null}.
 	 * @return the count of matching documents.
+	 * @since 3.4
 	 */
-	Mono<Long> count(Query query, Class<?> entityClass);
+	default Mono<Long> preciseCount(Query query, Class<?> entityClass) {
```
mp911de, line 956: How about `exactCount`? Precise implies a notion of number precision (single/double precision, decimals).

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
```diff
@@ -345,6 +347,47 @@ public void setEntityCallbacks(EntityCallbacks entityCallbacks) {
...
+
+	/**
+	 * En-/Disable usage of estimated count based on the given {@link BiPredicate estimationFilter}.
+	 *
+	 * @param enabled if {@literal true} {@link MongoCollection#estimatedDocumentCount()} will we used for {@link Document
+	 *          filter queries} that pass the given {@link BiPredicate estimationFilter}.
+	 * @param estimationFilter the {@link BiPredicate filter}.
+	 * @since 3.4
+	 */
+	public void useEstimatedCount(boolean enabled, BiPredicate<Document, CountOptions> estimationFilter) {
```
mp911de, line 369: For the start, I'd suggest making this method private and opening it up once we see demand for customization.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ReactiveMongoOperations.java
```diff
@@ -995,6 +1001,66 @@ default ReactiveSessionScoped withSession(Supplier<ClientSession> sessionProvide
...
+	 * This method may choose to use {@link #estimatedCount(Class)} for empty queries instead of running an
+	 * {@link com.mongodb.reactivestreams.client.MongoCollection#countDocuments(org.bson.conversions.Bson, com.mongodb.client.model.CountOptions)
+	 * aggregation execution} which may have an impact on performance.
+	 *
+	 * @param query the {@link Query} class that specifies the criteria used to find documents. Must not be
+	 *          {@literal null}.
+	 * @param entityClass class that determines the collection to use. Must not be {@literal null}.
+	 * @return the count of matching documents.
+	 */
+	Mono<Long> count(Query query, Class<?> entityClass);
```
mp911de, line 1025: Nit: We should probably slim down the second documentation paragraph and link to both count methods via `@see`.

mp911de (commented):
Did a first review pass and left a few comments up for discussion. The comments apply also to the reactive bits.

PR #3956, opened 2022-02-08
add comment geojsonpoint

sangyongchoi:
Hello, I found some discomfort while using GeoJsonPoint.
The meaning of x and y is not clear.
In my country, longitude and latitude are used more often than x and y.
I thought there would be a case like mine, so I added the following comment.
Please consider adding

thank you.


christophstrobl:
Thanks for bringing this up! We'll update the documentation to cover coordinate order/naming.

> If specifying latitude and longitude coordinates, list the longitude first and then latitude...

christophstrobl:
Thank you @sangyongchoi for your contribution! Merged to _main_ development line via bc027be.

PR #3982, opened 2022-03-02
Introduce support for property-based converters

christophstrobl:
This PR leverages the infrastructure provided by data-commons to  introduce conversions that target only specific properties of an entity. This allows a more fine grained control over the conversion and is an addition to the already existing reading & writing converter API.

Depends On: spring-projects/spring-data-commons#2566


mp911de (commented):
This is really neat and stores require a very slim implementation. Great work!

Requires a bit of reformatting.

mp911de:
That's merged, polished, and forward-ported now.

PR #4007, opened 2022-03-25
[Fix] Query with(pageable) and with(Pageable.unpaged)

Jatish-Khanna:


christophstrobl:
Thanks for the PR! Please spend some time describing the problem fix by the change. This helps not only us but also others who want to understand the reason behind it.

Jatish-Khanna:
There's a fix proposed when we want to use the same query for the pageable and unpaged instance. A typical case would be 

Get all the Object using the mongoTemplate from the Repository
      mongoTemplate.find(filterQuery.with(pageable), CustomBean.class)

Get the count of all the instance but unpaged
     mongoTemplate.count(filterQuery.with(Pageable.unpaged()), CustomBean.class);

And implementing the page instance
   new PageImpl<>(mongoDocumentsPaged, pageable, count);

In the order of invoking 
  with(pageable) comes first will set the parameters - "skip, limit, sort" and then invoking 
  with(Pageable.unpaged()) won't reset them to "0, 0, Sort.unsorted()" 

The fix will reset 

1) in the with(Pageable pageable)
  
			this.limit = 0; // default value
			this.skip = 0; // default value

2) in the with(Sort sort)
   this.sort = sort; // which will be Sort.unsorted()

Jatish-Khanna:
@christophstrobl please let me know if details to be added

Jatish-Khanna:
> 

I can understand that we can modify the Query with the static functions, although the feature has been provided to accept Pageable.unpaged() -> 

  Must auto-reset the properties "skip and limit" when an unpaged instance has been used as an argument.
 
  In the case mentioned, you must also reset the sort property of pagination via the static method which will be done automatically once unpaged instance has been used as argument.

ghost (approved):

PR #4015, opened 2022-04-04
Make sure to initialize PropvertyValueConversions in Converter setup.

christophstrobl:
Depending on: spring-projects/spring-data-commons#2591


mp911de:

PR #4016, opened 2022-04-05
Mark mongodb-driver-sync as non-optional

RasikaJoshiAM:
This package includes important class such as com.mongodb.client.MongoCollection;
and users of the client also needs to use this class, so this transitive dependency is important for downstream users as well. If we mark it as optional, then people will need to debug and manually include it. Ideally, we should give it to them.


christophstrobl:
Users may want to choose between the reactive and imperative driver bits and do not necessarily require the other one. There's also a note on this matter in the reference documentation.

PR #4059, opened 2022-05-20
Simplify usage of user provided aggregation operations.

christophstrobl:
Introduce `Aggregation.stage` which allows to use a plain JSON String or any valid `Bson` representation to be used within an aggregation pipeline stage, without having to implement `AggregationOperation` directly.
The change allows to make use of driver native builder API for aggregates.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
@@ -20,11 +20,15 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 
 import org.bson.Document;
+import org.bson.codecs.configuration.CodecRegistry;
 import org.springframework.beans.BeanUtils;
+import org.springframework.data.mongodb.CodecRegistryProvider;
 import org.springframework.data.mongodb.core.aggregation.ExposedFields.FieldReference;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ReflectionUtils;
+
+import com.mongodb.MongoClientSettings;
 
 /**
  * The context for an {@link AggregationOperation}.
@@ -33,7 +37,7 @@ import org.springframework.util.ReflectionUtils;
  * @author Christoph Strobl
  * @since 1.3
  */
-public interface AggregationOperationContext {
+public interface AggregationOperationContext extends CodecRegistryProvider {
 
 	/**
 	 * Returns the mapped {@link Document}, potentially converting the source considering mapping metadata etc.
@@ -113,5 +117,10 @@ public interface AggregationOperationContext {
 	 */
 	default AggregationOperationContext continueOnMissingFieldReference() {
 		return this;
+	}
+
+	@Override
+	default CodecRegistry getCodecRegistry() {
+		return MongoClientSettings.getDefaultCodecRegistry();
 	}
 }
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/BasicAggregationOperation.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/BasicAggregationOperation.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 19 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
```diff
@@ -114,4 +118,9 @@ default Fields getFields(Class<?> type) {
 	default AggregationOperationContext continueOnMissingFieldReference() {
 		return this;
 	}
+
+	@Override
+	default CodecRegistry getCodecRegistry() {
+		return MongoClientSettings.getDefaultCodecRegistry();
```
jyemin, line 124: In practice, when (if ever) will this default implementation be used?  The underlying driver never uses this method, because it doesn't take into account important user configuration (most notably, the UUID representation).

christophstrobl: Since it’s public API that users may have implemented, defaulting here eases the pain when upgrading to the new version. Internal implementations pick up the driver registry.
Are custom codecs heavily used?

jyemin: The main thing is that even the default registry is changed in the MongoClient constructor in order to apply the uuidRepresentation to the registry.  The default registry will throw when trying to encode a UUID value, because the default UuidRepresentation is UNSPECIFIED

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/BasicAggregationOperation.java
```diff
@@ -0,0 +1,61 @@
...
+			return document;
+		}
+
+		if (value instanceof Bson bson) {
+			return BsonUtils.asDocument(bson, context.getCodecRegistry());
+		}
+
+		if (value instanceof Map map) {
+			return new Document(map);
+		}
```
stIncMale, line 52: `BsonUtil.asMap` does not allow itself a liberty of just checking `instanceof Map`. I think that's because this check does not guarantee that the map keys are of the `String` type. For this reason, I don't think that
```java
if (value instanceof Map map) {
    return new Document(map);
}
```
is a type-safe code. Am I missing something?

Given the above, and that `BsonUtil.asMap` and `BsonUtil.asDocument` already check for special cases when some work may be skipped, can't the three `if` blocks be replaced with
```java
if (value instanceof Bson bson) {
    return BsonUtils.asDocument(bson, context.getCodecRegistry());
}
```
?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/util/BsonUtils.java
```diff
@@ -104,12 +123,26 @@ public static Map<String, Object> asMap(Bson bson) {
...
+	 * @since 4.0
+	 */
+	public static Document asDocument(Bson bson, CodecRegistry codecRegistry) {
+
+		if (bson instanceof Document document) {
+			return document;
 		}
 
-		Map<String, Object> map = asMap(bson);
+		Map<String, Object> map = asMap(bson, codecRegistry);
```
stIncMale, line 145: Given that `asMap` already does the `instanceof Document` check, this check may be safely removed here.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
stIncMale: Just a note: `MongoClientSettings.getDefaultCodecRegistry()` is also used explicitly in the default implementation of the `MongoConverter.getCodecRegistry` method.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/util/BsonUtils.java
```diff
@@ -81,18 +82,36 @@ public static <T> T get(Bson bson, String key) {
...
-			return ((DBObject) bson).toMap();
+		if (bson instanceof BasicDBObject dbo) {
+			return dbo;
+		}
+		if (bson instanceof DBObject dbo) {
+			return dbo.toMap();
 		}
 
-		return (Map) bson.toBsonDocument(Document.class, MongoClientSettings.getDefaultCodecRegistry());
+		return new Document((Map) bson.toBsonDocument(Document.class, codecRegistry));
```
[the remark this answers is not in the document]
christophstrobl: the additional `Document` wrapping can be removed, yes.
The `BasicDBObject` is still in there from 2.x generation and likely to be completely removed for 4.x.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
christophstrobl: true, same reason for those upgrading. As long as the user does not provide a custom implementation of the converter the codec registry (with all the user settings regarding UUID codec,...) will be used.

mp911de:

Issue #4123, opened 2022-07-20
 $looup does NOT work with @DBRef.


ArayChou:
I write a demo project to demostrate it. https://github.com/ArayChou/DBRefLookupTest . Please run unit test **test.QueryTest** to produce this issue.

`
@Document("person")
public class PersonPo {
    @Id
    private String id;
    private String name;
    @DBRef
    private PersonPo mother;
    @DocumentReference
    private PersonPo father;
}
`

`        AggregationOperation motherLookupDoesNotWork = Aggregation.lookup(
                "person",
                "mother.$id",
                "_id",
                "mother"
        );`

The above $lookup operation does NOT work. it should work. 

And I found a temporary solution, adding a (virtual)  field to the document as following, it works.
`      AddFieldsOperation addMotherId = Aggregation.addFields().addField("motherId")
                .withValue("$mother.$id")
                .build();
        AggregationOperation motherLookupWorks = Aggregation.lookup(
                "person",
                "motherId",
                "_id",
                "mother"
        );`

christophstrobl:
Thanks for reporting and the sample project!

PR #4125, opened 2022-07-20
Allow referencing the $id field of dbrefs within an aggregation pipeline

christophstrobl:
Closes: #4123


mp911de:

PR #4131, opened 2022-07-29
Add collation for an index via @CompoundIndex and @Index annotations.

### LINKED ISSUE #3002: @CompoundIndex has no configuration for collation and strength  [DATAMONGO-2133]
**[Bjorn Harvold](https://jira.spring.io/secure/ViewProfile.jspa?name=bjornharvold)** opened **[DATAMONGO-2133](https://jira.spring.io/browse/DATAMONGO-2133?redirect=false)** and commented

Request: Add support for collation and strength and other options to `@CompoundIndex`... or add documentation on adding all this is possible within def. E.g. def="\{a: 1, b:1}{collation:'en', strength:2}".

From what I can see, there is no support for adding collation or strength or similar options using `@CompoundIndex`. There might be a way to do it within the the "def" property but I cannot find any documentation that explains how to do this. Index fields and options are two different JSON pairs and I am not sure they would fit into "def" that only looks to cover the actual fields themselves

---

**Affects:** 2.1.2 (Lovelace SR2)

1 votes, 1 watchers

### LINKED ISSUE #4130: Allow to create case insensitive indexes via annotations
Hi,

the option for creating case insensitive index in MongoDB is available by setting `collation`. Specifically `locale` and `strength` values. It's described [in MongoDB documentation](https://www.mongodb.com/docs/manual/core/index-case-insensitive/).

Currently, annotations `@Indexed` and `@CompoundIndex` don't have `collation` in parameters to set.
Such option is available for `@Document` annotation but this will set value for all indexes in collection.

StefanTirea:
Closes #3002, closes #4130

Hi,
I recently wanted to use [Collation](https://www.mongodb.com/docs/manual/reference/collation/) with Spring Data and saw there is no support to implement it only with annotations for a specific field.
A use case would be to have an unique index which is case insensitive.
My PR adds support to use it with annotations (currently it is only possible programmatically).
I hope the implementation did not miss an obvious hurdle for this feature.

Still missing in this PR:
* Proper documentation in `@Index` & `@CompoundIndex` (see TODO)

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/index/CompoundIndex.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/index/CompoundIndex.java
@@ -46,6 +46,7 @@ import java.lang.annotation.Target;
  * @author Johno Crawford
  * @author Christoph Strobl
  * @author Dave Perryman
+ * @author Stefan Tirea
  */
 @Target({ ElementType.TYPE })
 @Documented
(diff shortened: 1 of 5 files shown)


christophstrobl:
Thank you @StefanTirea! Merged to _main_.

Issue #4132, opened 2022-08-02
MongoTemplate.aggregateStream does not support local variables


Saljack:
Hello 
I found an issue in `aggregateStream` method on `MongoTemplate`. If you use a local variable (`$$variable_name`) for example in filter operator then it is not able to resolve this local variable name and throws:
```
org.springframework.data.mapping.context.InvalidPersistentPropertyPath: No property '$$localvar' found on class ***.TraceEntity! Did you mean: ?
	at org.springframework.data.mapping.context.PersistentPropertyPathFactory.createPersistentPropertyPath(PersistentPropertyPathFactory.java:205)
	at org.springframework.data.mapping.context.PersistentPropertyPathFactory.lambda$getPersistentPropertyPath$1(PersistentPropertyPathFactory.java:172)
...
```
Everything works fine with the `aggregate` method. I think the issue is in the aggregation context because `aggregateStream` uses `TypeBasedAggregationOperationContext` and `aggregate` uses `RelaxedTypeBasedAggregationOperationContext`. Which is weird I would excepted the same aggregation context type.

Here is an example:
### MyEntity:
```json
{
  "activitaTrace" : [
     {"type": "USAGE_CLOUD"},
     {"type": "OTHER_USAGE"}
  ]
}
```

### Aggregation
```java
ProjectionOperation projection = project()
    .and(
        filter("activityTrace")
            .as("localvar")
            .by(
                Eq.valueOf("$$localvar.type")
                    .equalToValue("USAGE_CLOUD")))
    .as("filtered");

TypedAggregation<UsageTracking> aggregation = newAggregation(UsageTracking.class, List.of(projection));
List<Document> aggregate = mongoTemplate.aggregate(aggregation, Document.class).getMappedResults(); // It works

List<Document> aggregateStream = mongoTemplate.aggregateStream(aggregation, Document.class).stream().toList(); // throws InvalidPersistentPropertyPath
```

I use spring-data-mongodb `3.3.4` but I also checked the latest version (`4.0.0-SNAPSHOT`) and there is also the same aggregation context in the `aggregateStream`. So I assume the latest version is also affected.

christophstrobl:
@Saljack thanks for reporting!

PR #4147, opened 2022-08-23
Favor relaxed type mapping over strict one for aggregateStream.

christophstrobl:
Closes: #4132


mp911de:

Issue #3187, opened 2019-08-01
Add support for Aggregations in Update. [DATAMONGO-2331]


spring-projects-issues:
**[Christoph Strobl](https://jira.spring.io/secure/ViewProfile.jspa?name=cstrobl)** opened **[DATAMONGO-2331](https://jira.spring.io/browse/DATAMONGO-2331?redirect=false)** and commented

---

**Reference URL:** https://docs.mongodb.com/manual/reference/method/db.collection.update/#update-with-an-aggregation-pipeline

**Issue Links:**
- [DATAMONGO-2363](https://jira.spring.io/browse/DATAMONGO-2363) Add support for $merge aggregation stage
 (_**"is depended on by"**_)

**Referenced from:** pull request https://github.com/spring-projects/spring-data-mongodb/pull/789, and commits https://github.com/spring-projects/spring-data-mongodb/commit/0cc3532f0832c4f0ba51e2cc0d829836dfe0e138, https://github.com/spring-projects/spring-data-mongodb/commit/07b52d08b22e2291d5206b637daf36bff8f40828, https://github.com/spring-projects/spring-data-mongodb/commit/1a5de2e1db939f7b35579f11815894fd637fc227, https://github.com/spring-projects/spring-data-mongodb/commit/476c758982b5443bfd83fc1704c5aa5852c46832, https://github.com/spring-projects/spring-data-mongodb/commit/cc07a1bbb483885f9268351d4588b1f0722908d5, https://github.com/spring-projects/spring-data-mongodb/commit/46b038c20040bc59c4f943b48d02794f093debf9, https://github.com/spring-projects/spring-data-mongodb/commit/afa0b0a95ef7d5a4dccfcb78e2b797bbf449fc01, https://github.com/spring-projects/spring-data-mongodb/commit/32cbae0e5f813fb662f8fb0f54d9c642f25d983d

Issue #2602, opened 2017-05-10
Provide Kotlin extensions for Class based methods in MongoOperations / ReactiveMongoOperations [DATAMONGO-1689]


spring-projects-issues:
**[Sébastien Deleuze](https://jira.spring.io/secure/ViewProfile.jspa?name=sdeleuze)** opened **[DATAMONGO-1689](https://jira.spring.io/browse/DATAMONGO-1689?redirect=false)** and commented

Using what has been done in [MiXiT](https://github.com/mixitconf/mixit/blob/master/src/main/kotlin/mixit/util/Extensions.kt#L37-L54) project as a starting point, the purpose of this issue is provide builtin Kotlin extensions leveraging [Kotlin reified type parameters](https://kotlinlang.org/docs/reference/inline-functions.html) to avoid specifying explicitly types to serialize/deserialize in APIs like `MongoOperations` and `ReactiveMongoOperations`. Providing `KClass` based variants should also be useful.

I am working on a PR I will submit asap finished

---

**Issue Links:**
- [DATACASS-484](https://jira.spring.io/browse/DATACASS-484) Add Kotlin extensions for Template API
 (_**"Cloned to"**_)

**Referenced from:** pull request https://github.com/spring-projects/spring-data-mongodb/pull/463

PR #789, opened 2019-09-17
DATAMONGO-2331 - Add support for Update with an aggregation pipeline.

christophstrobl:
⚠️ **tentative feature for 2.3** ⚠️ 

---

Now the _findAndModify_ & _update_ methods exposed by (`Reactive`)`MongoOperations` also accept an [Aggregation Pipeline](https://docs.mongodb.com/manual/release-notes/4.2/#update-and-aggregation) via `AggregationUpdate`.

The update can consist of the following stages:

* `AggregationUpdate.set(...).toValue(...)` -> `$set : { ... }`
* `AggregationUpdate.unset(...)` -> `$unset : [ ... ]`
* `AggregationUpdate.replaceWith(...)` -> `$replaceWith : { ... }`

```java
AggregationUpdate update = Aggregation.newUpdate()
    .set("average").toValue(ArithmeticOperators.valueOf("tests").avg())
    .set("grade").toValue(ConditionalOperators.switchCases(
        when(valueOf("average").greaterThanEqualToValue(90)).then("A"),
        when(valueOf("average").greaterThanEqualToValue(80)).then("B"),
        when(valueOf("average").greaterThanEqualToValue(70)).then("C"),
        when(valueOf("average").greaterThanEqualToValue(60)).then("D"))
        .defaultTo("F")
    );

template.update(Student.class)
    .apply(update)
    .all();
```

The snippet above triggers the following 

```javascript
db.students.update(
   { },
   [
     { $set: { average : { $avg: "$tests" } } }, 
     { $set: { grade: { $switch: { 
                           branches: [
                               { case: { $gte: [ "$average", 90 ] }, then: "A" },
                               { case: { $gte: [ "$average", 80 ] }, then: "B" },
                               { case: { $gte: [ "$average", 70 ] }, then: "C" },
                               { case: { $gte: [ "$average", 60 ] }, then: "D" }
                           ],
                           default: "F"
     } } } }
   ],
   { multi: true }
)
```

PR #4154, opened 2022-09-05
Align signature of Kotlin extension functions to match Java API.

### LINKED ISSUE #4153: Switch MongoOperations.updateMulti Kotlin extension function to accept a more generic UpdateDefinition
Today I was surprised, that I can't express the same update with Kotlin, which I can express with Java's MongoOperations API. My code was
```java
mongoTemplate.updateMulti(
    Query.query(Criteria.where("someField").type(JsonSchemaObject.Type.DOUBLE)),
    AggregationUpdate.update()
        .set(...),
  MyEnt::class.java
)
```

Usually Kotlin extension allows me to write the same code, but without ugly `::class.java`. So I would expect the following to work:

```kotlin
mongoTemplate.updateMulti<MyEnt>(
    Query.query(Criteria.where("someField").type(JsonSchemaObject.Type.DOUBLE)),
    AggregationUpdate.update()
        .set(...)
)
```
but it does not compile.

The reason is that the signature of Kotlin extension for `MongoOperations` is different from the one defined on the original `MongoOperations` interface method:

https://github.com/spring-projects/spring-data-mongodb/blob/3342c75ecf3e560c70637155df8cae658ff4a6bb/spring-data-mongodb/src/main/kotlin/org/springframework/data/mongodb/core/MongoOperationsExtensions.kt#L538-L540

https://github.com/spring-projects/spring-data-mongodb/blob/3342c75ecf3e560c70637155df8cae658ff4a6bb/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoOperations.java#L1511

The Kotlin extension accepts a more narrow type `org.springframework.data.mongodb.core.query.Update` type, while the Java interface accepts a more generic `org.springframework.data.mongodb.core.query.UpdateDefinition`.

By looking at the implementation of the Kotlin extension, I see not reasons for it to accept a different type. So, I think the type accepted by Kotlin extension function should be aligned with the one defined on the original interface it extends.

gavvvr:
Closes #4153

Caused by: #789 (Kotlin extensions were not adjusted accordingly)

Related tickets: #2602 #3187

### DIFF
--- a/spring-data-mongodb/src/main/kotlin/org/springframework/data/mongodb/core/MongoOperationsExtensions.kt
+++ b/spring-data-mongodb/src/main/kotlin/org/springframework/data/mongodb/core/MongoOperationsExtensions.kt
@@ -28,7 +28,7 @@ import org.springframework.data.mongodb.core.mapreduce.MapReduceOptions
 import org.springframework.data.mongodb.core.mapreduce.MapReduceResults
 import org.springframework.data.mongodb.core.query.NearQuery
 import org.springframework.data.mongodb.core.query.Query
-import org.springframework.data.mongodb.core.query.Update
+import org.springframework.data.mongodb.core.query.UpdateDefinition
 import java.util.stream.Stream
 
 /**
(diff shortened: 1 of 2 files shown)


christophstrobl:
@gavvvr thank you for the contribution - merged to _main_.

Issue #4104, opened 2022-07-08
GeoJsonPolygonconvertor does not convert inner rings.


lennertdefeyter:
Hi,

I noticed that the GeojsonPolygonconvertor does not convert the inner rings. If you want to convert a polygon with holes. It only looks at the outer ring.

This is very apparent in the code in line 759 of GeoConverters.java.
`static GeoJsonPolygon toGeoJsonPolygon(List dbList) {
		return new GeoJsonPolygon(toListOfPoint((List) dbList.get(0)));
	}`

As you can see, it only looks at the first item in the list of rings (e.g. the outer ring).

I've already fixed it with a custom convertor (see attachment, although this is for multipolygons), but it might be nice to have this in the core.

I think that in order to fix it, the toGeoJsonPolygon(List dbList) function in GeoConverters.java needs to change to:

`
	static GeoJsonPolygon toGeoJsonPolygon(List dbList) {
		Iterator<GeoJsonLineString> it = ((List) dbList).iterator();
		GeoJsonPolygon g = new GeoJsonPolygon(toListOfPoint((List) it.next()));
		while (it.hasNext())
			g = g.withInnerRing(toListOfPoint((List) it.next()));
		return g;
	}`

[GeoJsonMultiPolygonDeserializer.java.txt](https://github.com/spring-projects/spring-data-mongodb/files/9069627/GeoJsonMultiPolygonDeserializer.java.txt)
.

spring-projects-issues:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

lennertdefeyter:
Hi @christophstrobl ,

Sorry for the delay. This is version 2.7.2.

Attached is a zip that should show the problem.

it inserts a polygon correctly with a donut hole in the center: outer ring is a box with bounds: 0,0 -> 10,10 . Inner ring is a box 4,4 -> 6,6

It is stored correctly in the database. But when querying it clearly has only the outer ring:
![image](https://user-images.githubusercontent.com/60380125/181593285-c53a8e79-e03b-4636-9f7f-7d0a50918658.png)

[mongodbdemo.zip](https://github.com/spring-projects/spring-data-mongodb/files/9212676/mongodbdemo.zip)

Thank you for your help.

christophstrobl:
Sorry for long silence - good catch! Thank you for the reproducer!

mp911de:
[...]

PR #4156, opened 2022-09-08
Fix GeoJson polygon conversion for polygons with inner ring.

christophstrobl:
Closes: #4104


mp911de:

PR #4160, opened 2022-09-12
Replace deprecated StreamUtils API

bclozel:
As of spring-projects/spring-framework#29125,
`StreamUtils..emptyInput()` is deprecated in favor of
`InputStream.nullInputStream()` from the JDK.


christophstrobl:
thanks @bclozel !

Issue #4167, opened 2022-09-16
Change stream option startAfter not applied via reactive template.


christophstrobl:
The `startAfter` option is incorrectly applied as `resumeAfter`.

Related to: #739

PR #4168, opened 2022-09-16
Fix usage of change stream option startAfter. 

christophstrobl:
closes: #4167


mp911de:

PR #4181, opened 2022-09-25
Support automatic registration of converters

sangyongchoi:
Hi, i am spring-data-mongo user.
I had some inconvenience when registered converters.
Is currently manually register converters.
```kotlin
@Bean
fun mongoCustomConversions(): MongoCustomConversions {
    val converters = applicationContext.getBeansOfType(Converter::class.java).values
        .stream().collect(Collectors.toList())
    return MongoCustomConversions(converters)
}
```
Can I change to automatic registration from manually registration ?
If it's not possible, can you tell me why?

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
@@ -161,6 +161,7 @@ import com.mongodb.client.result.UpdateResult;
  * @author Yadhukrishna S Pai
  * @author Anton Barkan
  * @author Bartłomiej Mazur
+ * @author Sangyong Choi
  */
 public class MongoTemplate implements MongoOperations, ApplicationContextAware, IndexOperationsProvider {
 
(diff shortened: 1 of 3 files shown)


mp911de:
How is that supposed to work if a user has multiple Spring Data modules on the class path. We would register all converters in all modules without having a way to distinguish whether a `Converter` is intended for Spring Data usage and for which module it should be used.

sangyongchoi:
> How is that supposed to work if a user has multiple Spring Data modules on the class path. We would register all converters in all modules without having a way to distinguish whether a `Converter` is intended for Spring Data usage and for which module it should be used.

Thank you for reply 😄 
So i hope add interface (e.g MongoConverter).
Then, this is possible as follows.
```kotlin
if (listableBeanFactory != null) {
	converters = listableBeanFactory
			.getBeansOfType(MongoConverter.class)
			.values()
			.stream()
			.toList();
} else {
	converters = Collections.emptyList();
}
```

Is this possible?

mp911de:
Thanks for your suggestion. We want to keep the existing mechanism and keep the converter registration as simple as possible. If adding an interface to your converters works in your project, then we encourage you to follow that path and apply that pattern in your project.

Issue #4139, opened 2022-08-17
Add support for MongoDB 6.0 Aggregation changes


mp911de:
See https://www.mongodb.com/docs/manual/release-notes/6.0/#aggregation-stages-and-operators and #4138

PR #4182, opened 2022-09-26
Add missing aggregation operators and stages

christophstrobl:
Closes: #4139


mp911de:
I like how this pull request shapes up, especially the interop across temporal units. There are some operators that require code reformatting and some public methods without Javadoc.

mp911de:

PR #4196, opened 2022-10-09
update Type "boolean" to "bool"

gongxuanzhang:
mongodb command :  {$type}  
When you use boolean match，you will have a error like : Query failed with error code 2 and error message 'Unknown type name alias: boolean'

because mongodb manual {$type : "bool" } , not "boolean";
link [mongodb type desc](https://www.mongodb.com/docs/manual/reference/operator/query/type/ )


christophstrobl:
Good catch - thanks for the PR.

Issue #4184, opened 2022-09-27
Failed usage of provided ID value on document insert


aburmakov:
Self-generated ID cannot be insert into mongo collection. System-generated ObjectID.get() is used instead of self-generatd.
Looks like problem appears after commit: 8ee33b2e5 ("Generate and convert id on insert if explicitly defined.", 2022-05-19)

**Sample code:**
```java
@Document(collection = AcmeDocument.COLLECTION_NAME)
public class AcmeDocument implements IdDocument {
    public static final String COLLECTION_NAME = "acme";
    @MongoId
    private String id;
    private String userName;
    public AcmeDocument(String userName) {
        this.id = UUID.randomUUID().toString();
        this.userName = userName;
    }
//...
}

// usage:
AcmeDocument d = new AcmeDocument(userName);
d = acmeRepository.save(d);
log.debug("create.id: {}", d.getId());
```

```
Log output: create.id: e43afa9b-06db-4935-a94a-f76578c7af83
But mongoDb document has _Id as generated ObjectId: 6332ca654af6176f2c0c230d
```

pcornelissen:
save also creates the wrong ID in spring boot 2.7.4: https://github.com/pcornelissen/springboot-mongo-bug 

@mp911de As you have worked on that, is there an ETA for a new release?

Issue #4197, opened 2022-10-10
lost field value annotated with @MongoId after upgrade springboot from 2.7.0 to 2.7.4


wujimin:
# problem
```
public class Model {
  @MongoId
  private String myId;
}

template.insert(new Model().setMyId("abc");
```
the value of "_id" saved to mongo:  
* before upgrade: "abc"  
* after upgrade: a new generated value

# Possible Solutions
https://github.com/spring-projects/spring-data-mongodb/blob/a5725806f5cf375dfc5343bfc2cda223ae97e5ab/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/QueryOperations.java#L275 

maybe should change this line to 
`if (entity == null || source.hasId()) { `

wujimin:
seems same to https://github.com/spring-projects/spring-data-mongodb/issues/4184

wujimin:
should invoke `insert`, not `save`

christophstrobl:
ah - thank you very much @wujimin !

wujimin:
you are welcome~

PR #4203, opened 2022-10-12
Preserve given Id on insert. 

christophstrobl:
Closes: #4184 & #4197

needs to be back ported.


mp911de:

PR #4216, opened 2022-10-20
Improve configuration support for Observability integration

gregturn:
CC @marcingrzejszczak


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoMetricsConfigurationHelper.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.mongodb.client.SynchronousContextProvider;
+
+/**
+ * Helper functions to ease registration of Spring Data MongoDB's observability.
+ */
+public class MongoMetricsConfigurationHelper {
+
+	public static SynchronousContextProvider synchronousContextProvider(Tracer tracer, ObservationRegistry registry) {
+		return () -> new SynchronousTraceRequestContext(tracer).withObservation(Observation.start("name", registry));
```
gregturn, line 15: If we don't have an `Observation.start()` somewhere, then nothing seems to get logged in Zipkin. This is what I had used in the past inside the ZipkinIntegrationTests, but it feels quite arbitarary.

If there is a better place to "start" things, I would like to know and could adjust things accordingly.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoMetricsReactiveConfigurationHelper.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Helper functions to ease registration of Spring Data MongoDB's observability.
+ */
+public class MongoMetricsReactiveConfigurationHelper {
+
+	public static ReactiveContextProvider reactiveContextProvider(ObservationRegistry registry) {
+		return subscriber -> {
+			if (subscriber instanceof CoreSubscriber<?> coreSubscriber) {
+				return new ReactiveTraceRequestContext(coreSubscriber.currentContext())
+						.withObservation(Observation.start("name", registry));
```
gregturn, line 19: Same comment as https://github.com/spring-projects/spring-data-mongodb/pull/4216/files#r1000663761

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoTracingObservationHandler.java
```diff
@@ -49,6 +50,10 @@ public MongoTracingObservationHandler(Tracer tracer) {
 		this.tracer = tracer;
 	}
 
+	public void register(ObservationRegistry observationRegistry) {
```
gregturn, line 53: This shortcut method was purely created to slim down the registration statement users would have to add to their own code.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/ContextProviderFactory.java
```diff
@@ -0,0 +1,139 @@
...
+		INSTANCE;
+
+		@Override
+		public RequestContext getContext(Subscriber<?> subscriber) {
+
+			if (subscriber instanceof CoreSubscriber<?> cs) {
+
+				Map<Object, Object> map = cs.currentContext().stream()
+						.collect(Collectors.toConcurrentMap(Entry::getKey, Entry::getValue));
+				if (map.containsKey(ObservationThreadLocalAccessor.KEY)) {
```
marcingrzejszczak, line 108: Why do we then store or under a different key instead of referencing it from that key later on?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/DefaultMongoHandlerObservationConvention.java
```diff
@@ -15,34 +15,75 @@
...
+
 /**
  * Default {@link MongoHandlerObservationConvention} implementation.
  *
  * @author Greg Turnquist
- * @since 4.0.0
+ * @since 4
  */
-public class DefaultMongoHandlerObservationConvention implements MongoHandlerObservationConvention {
+class DefaultMongoHandlerObservationConvention implements MongoHandlerObservationConvention {
```
marcingrzejszczak, line 40: Don't we want this to be public so that the users can extend this and add more things that they want?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/DefaultMongoHandlerObservationConvention.java
```diff
@@ -15,34 +15,75 @@
...
 		}
 
+		ServerAddress serverAddress = context.getCommandStartedEvent().getConnectionDescription().getServerAddress();
+
+		if (serverAddress != null) {
+
+			keyValues = keyValues.and(LowCardinalityCommandKeyNames.NET_TRANSPORT.withValue("IP.TCP"),
+					LowCardinalityCommandKeyNames.NET_PEER_ADDR.withValue(serverAddress.getHost()));
+
+			InetSocketAddress socketAddress = serverAddress.getSocketAddress();
```
marcingrzejszczak, line 77: I don't remember but wasn't one of this calls very slow?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/DefaultMongoHandlerObservationConvention.java
```diff
@@ -54,8 +95,12 @@ public KeyValues getLowCardinalityKeyValues(MongoHandlerContext context) {
...
 	public KeyValues getHighCardinalityKeyValues(MongoHandlerContext context) {
 
-		return KeyValues.of(
-				HighCardinalityCommandKeyNames.MONGODB_COMMAND.withValue(context.getCommandStartedEvent().getCommandName()));
+		return KeyValues.of(HighCardinalityCommandKeyNames.MONGODB_COMMAND.withValue(context.getCommandName()));
+	}
+
+	@Override
+	public String getContextualName(MongoHandlerContext context) {
+		return context.getContextualName();
```
marcingrzejszczak, line 103: It would be better to just define the contextual name here and not take it from the context.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoHandlerContext.java
```diff
@@ -25,37 +23,49 @@
...
  * A {@link Observation.Context} that contains MongoDB events.
  *
  * @author Marcin Grzejszczak
  * @author Greg Turnquist
- * @since 4.0.0
+ * @author Mark Paluch
+ * @since 4.0
  */
-public class MongoHandlerContext extends Observation.Context {
+class MongoHandlerContext extends SenderContext<Object> {
```
marcingrzejszczak, line 44: If this is not public users won't be able to access it from the convention

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoObservationCommandListener.java
```diff
@@ -75,11 +81,26 @@ public void commandStarted(CommandStartedEvent event) {
...
+
+		if (parent != null) {
+			observation.parentObservation(parent);
 		}
 
-		setupObservability(event, requestContext);
+		observation.start();
+
+		requestContext.put(Observation.class, observation);
+		requestContext.put(MongoHandlerContext.class, observationContext);
```
marcingrzejszczak, line 98: You don't need to store the context anymore cause you can retrieve it from an observation

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/observability/ImperativeIntegrationTests.java
```diff
@@ -0,0 +1,77 @@
...
+			repository.deleteAll();
+			repository.save(new Person("Dave", "Matthews", 42));
+			List<Person> people = repository.findByLastname("Matthews");
+
+			assertThat(people).hasSize(1);
+			assertThat(people.get(0)).extracting("firstname", "lastname").containsExactly("Dave", "Matthews");
+
+			repository.deleteAll();
+
+			System.out.println(((SimpleMeterRegistry) meterRegistry).getMetersAsString());
```
marcingrzejszczak, line 74: You can do assertions on spans and metrics if you want to

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/DefaultMongoHandlerObservationConvention.java
mp911de: We can open up the code later on if someone really wants to build on our implementation.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/DefaultMongoHandlerObservationConvention.java
mp911de: Socket addresses must be resolved at this time already, so we use a cached instance.

PR #4218, opened 2022-10-21
Aligns the context entries with the rest of the portfolio

marcingrzejszczak:
without this change each project will set the observation in the reactive context to `micrometer.observation` and only for mongodb will we need to set it to `Observation.class`. This has to be changed to be consistent.


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoObservationCommandListener.java
```diff
@@ -114,8 +114,7 @@ public void commandStarted(CommandStartedEvent event) {
 
 		observation.start();
 
-		requestContext.put(Observation.class, observation);
-		requestContext.put(MongoHandlerContext.class, observationContext);
+		requestContext.put("micrometer.observation", observation);
```
mp911de, line 117: Can we use constants instead of text keys? The other changes seem fine.

mp911de (changes requested):

Issue #4244, opened 2022-12-02
Ahead Of Time support for Querydsl


susimsek:
hi spring graphql teams.i am using spring graphql with querydsl. When I compile the project as native, I am getting an error as follows. Can you provide aot support for Querydsl?
<img width="1597" alt="image" src="https://user-images.githubusercontent.com/46258776/205240975-8df9d4f2-bd72-4620-bb4b-97aa0355f96d.png">

The reproduce example codes are shared via my Github, check [here.](https://github.com/susimsek/spring-graphql-samples) 

Spring boot version: 3.0.0
 Kotlin version: 1.7.21
 Java version: 17

christophstrobl:
Thanks for reporting and the reproducer - we'll take care of it!
Please do not upload images of error messages but use code fences instead. Thank you.
```
Caused by: java.lang.IllegalStateException: No suitable constructor found on class org.springframework.data.mongodb.repository.support.ReactiveQuerydsIMongoPredicateExecutor to match the given arguments: org.springframework.data.mongodb.repository.support.MappingMongoEntityInformation,org.springframework.data.mongodb.core.ReactiveMongoTemplate.
Make sure you implement a constructor taking these
at org.springframework.data.repository.core.support.RepositoryFactorySupport.lambda$instantiateClass$6(RepositoryFactorySupport.java:572)~[spring-graphql-samples:3.0.0]
at java.base@17.0.5/java.util.Optional.orElseThrow(Optional.java:403)~[spring-graphql-samples:na]
at org.springframework.data.repository.core.support.RepositoryFactorySupport.instantiateClass(RepositoryFactorySupport.java:572)
~[spring-graphql-samples:3.0.0]
at org.springframework.data.mongodb.repository.support.ReactiveMongoRepositoryFactory.getRepositoryFragments(ReactiveMongoRepositoryFactory.java:106)
~[na:na]
at org.springframework.data.repository.core.support.RepositoryFactorySupport.getRepositoryComposition(RepositoryFactorySupport.java:428)~[spring-graphql-samples:3.0.0]
at org.springframework.data.repository.core.support.RepositoryFactorySupport.getRepository(RepositoryFactorySupport.java:293)
~[spring-graphql-samples:3.0.0]
at org. springframework.data.repository.core.support.RepositoryFactoryBeanSupport.lambda$afterPropertiesSet$5(RepositoryFactoryBeanSupport.java:279)~[spring-graphql-samples:3.0.0]
[...]

PR #4245, opened 2022-12-02
Add missing hints for Querydsl integration

christophstrobl:
This PR adds missing reflection configuration for Querydsl integration. 
We now also make sure to call the `queryMixing` getter instead of reading the field via reflection.

Manual registration of `Q` classes is still required (see: spring-projects/spring-data-commons#2721)

Closes: #4244 

---

Should be back ported to _4.0.x_


mp911de:

Issue #4284, opened 2023-02-01
Support property based explicit field encryption


christophstrobl:
Extend the property based conversion feature to combine `@Encrypted` with `@ValueConverter` into a single annotation that is evaluated when mapping domain types into their store native representation and back.

```java
@Encrypted
@ValueConverter(MongoCryptConverter.class)
@interface EncryptedField {

        @AliasFor(annotation = ValueConverter.class, value = "value")
        Class<? extends PropertyValueConverter> converter() default MongoCryptConverter.class;

	@AliasFor(annotation = Encrypted.class, value = "algorithm")
	String algorithm() default "";

	String altKeyName() default "";
}
```

The `EncryptingConverter` SPI offers dedicated methods to en-/decrypt objects using an `EncryptionContext` that provides access to the underlying encryption engine plus all the additional information derived from the target property itself.

```java
interface EncryptingConverter implements MongoValueConverter<Object, Object> {

	@Override
	default Object read(Object value, MongoConversionContext context) {
	    return decrypt(value, buildEncryptionContext(context));
	}	

	@Override
	default BsonBinary write(Object value, MongoConversionContext context) {
		return encrypt(value, buildEncryptionContext(context));
	}

	Object decrypt(Object value, MongoEncryptionContext context);

	BsonBinary encrypt(Object value, MongoEncryptionContext context)
	
	MongoEncryptionContext buildEncryptionContext(MongoConversionContext context);
}
```

The default implementation (`MongoCryptConverter implements EncryptingConverter`) will be using `com.mongodb.client.vault.ClientEncryption` for en-/decrypt operations and is set as default for `@EncryptedField(converter)`.

To use explicit encryption in a reactive setup one can choose to use the imperative converter (still using the sync mongodb encryption library) within the reactive flow or switch the default converter to a fully reactive variant (using the drivers reactive streams API) resolving values on subscribe.

```java
interface ReactiveEncryptingConverter extends MongoValueConverter<Object, Object> {

	@Override
	default Object read(Object value, MongoConversionContext context) {
	    return decrypt(value, buildEncryptionContext(context));
	}	

	@Override
	default BsonBinary write(Object value, MongoConversionContext context) {
		return encrypt(value, buildEncryptionContext(context));
	}

	Mono<Object> decrypt(Object value, MongoEncryptionContext context);

	Mono<BsonBinary> encrypt(Object value, MongoEncryptionContext context)

[...]

PR #4302, opened 2023-02-22
Add support for explicit field encryption.

christophstrobl:
Still a **draft**.

Closes: #4284


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/encryption/EncryptionKey.java
```diff
@@ -0,0 +1,157 @@
...
+			return Type.ID;
+		}
+
+		@Override
+		public String toString() {
+
+			if (BsonBinarySubType.isUuid(value.getType())) {
+				String representation = value.asUuid().toString();
+				if (representation.length() > 6) {
+					return String.format("KeyId('%s***')", representation.substring(0, 6));
```
mp911de, line 77: I have the feeling that we should avoid rendering the key, even subparts in `toString`. We could rather replace each character with a `*` and retain the dashes if we wanted to.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/encryption/EncryptionKey.java
```diff
@@ -0,0 +1,157 @@
...
+			return Type.ALT;
+		}
+
+		@Override
+		public String toString() {
+
+			if (value().length() <= 3) {
+				return "AltKeyName('***')";
+			}
+			return String.format("AltKeyName('%s***')", value.substring(0, 3));
```
mp911de, line 119: Same topic about rendering the key name like `KeyId.toString()`

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/encryption/MongoClientEncryption.java
```diff
@@ -0,0 +1,113 @@
...
+			public boolean refresh() {
+				return false;
+			}
+		};
+	}
+
+	/**
+	 * @return {@literal true} if refreshed, {@literal false} otherwise.
+	 */
+	public boolean refresh() {
```
mp911de, line 78: I'm wondering whether we should rather show an `Lazy` example for singleton usage and always call `Supplier.get()`.

mp911de (commented):
Left a few comments from an initial review. I like the direction these changes took.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/encryption/EncryptingConverter.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.data.mongodb.core.convert.MongoConversionContext;
+import org.springframework.data.mongodb.core.convert.MongoValueConverter;
+
+/**
+ * A specialized {@link MongoValueConverter} for {@literal en-/decrypting} properties.
+ *
+ * @author Christoph Strobl
+ * @since 4.1
+ */
+public interface EncryptingConverter<S, T> extends MongoValueConverter<S, T> {
```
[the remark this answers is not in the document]
christophstrobl: valid point, we need to untangle bits here to avoid cycle between the `convert` and `encrypt` package. Basically we need to make sure `EncryptionContext` is not a `ValueConversionContext` any longer

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/encryption/MongoClientEncryption.java
christophstrobl: I had thought about using `Lazy` but decided against it to have more control over the refresh. Also calling the supplier may lead to unnecessary creation of `ClientEncryption` instances that all hold connections to the database.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/encryption/EncryptionKey.java
christophstrobl: I had looked at `EncryptionKey` and `KeyImpl` from `javax.security.auth.kerberos` which do render the bits right away. `DESKey` does not. `P11Key` tells you the type and length. So I'm not sure. Maybe @rwinch has an opinion.

mp911de:

Issue #3218, opened 2019-09-12
MongoTemplate should add query hint when executing update [DATAMONGO-2362]


spring-projects-issues:
**[Christoph Strobl](https://jira.spring.io/secure/ViewProfile.jspa?name=cstrobl)** opened **[DATAMONGO-2362](https://jira.spring.io/browse/DATAMONGO-2362?redirect=false)** and commented

---

**Reference URL:** https://docs.mongodb.com/manual/release-notes/4.2/#update-and-hint

PR #4311, opened 2023-03-02
Support hints on Update.

christophstrobl:
This PR makes sure to read `Query` hints and apply them to the MongoDB `UpdateOptions` when running an _update _via `Reactive-/MongoTemplate`.

Closes: #3218


mp911de:

PR #4315, opened 2023-03-03
Migrates DependencyTests to ArchUnit.

schauder:
Migrates DependencyTests to ArchUnit.

The tests are currently failing due to various undesired dependencies.

The tests start with something like:

		JavaClasses importedClasses = new ClassFileImporter()
		        ...
				.that(onlySpringData())
				
By adding

				.that(ignorePackage("org.springframework.data.aot.hint"))

or 

				.that(ignore(SomeClass.class));
				
the given classes or packages can be ignored for the analysis.
Of course the goal should be to fix the cycles.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.1.0-SNAPSHOT</version>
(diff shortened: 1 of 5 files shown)


christophstrobl:
I'm curious what's your take on the cycles created by the `config`, `support` and `util` packages?
The AOT ones I'd rather skip for now.

PR #4317, opened 2023-03-06
Add support for Keyset- and Offset-based cursor scrolling

### LINKED ISSUE #4308: Add support for keyset scrolling
Based on https://github.com/spring-projects/spring-data-commons/issues/2151, we want to add means to consume a query result set through a scroll API to be able to resume querying at a later stage. We want to introduce keyset-based pagination which allows using keysets to reconstruct a scroll position for subsequent queries.

Tasks:
- [ ] #4317
- [x] #4325
- [x] #4326

mp911de:
Usage examples:

**Repository query methods**

```java

interface UserRepository extends Repository<User, Long> {

  Scroll<User> findFirst10ByLastnameOrderByFirstname(String lastname, OffsetScrollPosition position);
}

Scroll<User> users = repository.findFirst10ByLastnameOrderByFirstname("Doe", OffsetScrollPosition.initial());

do {

  for (User u : users) {
    // consume the user
  }

  // obtain the next Scroll
  users = repository.findFirst10ByLastnameOrderByFirstname("Doe", users.lastScrollPosition());
} while (!users.isEmpty() && !users.isLast());
```

**Template API**

```java
MongoTemplate template = …;

Query q = new Query(where("firstName").regex("J.*"))
              .with(Sort.by("firstName", "age"))
              .limit(2)
              .with(KeysetScrollPosition.initial());

Scroll<Person> scroll = template.scroll(q, Person.class);

Scroll<Person> scroll = template.query(Person.class)
                               .matching(Query.query(where("firstName").regex("J.*"))
                                              .with(Sort.by("firstName", "age")))
                               .scroll(OffsetScrollPosition.of(10));
```

Tasks:

- [x] Implementation
- [x] Revisit `Sort` precedence and limit/skip relationship to `CursorRequest` for methods accepting `Query` and `CursorRequest` types
- [x] Documentation

Limitations:

- Using projections with Keyset-scrolling requires all keyset properties to be present in the target projection type

Closes #4308 
Depends on https://github.com/spring-projects/spring-data-commons/pull/2787

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
@@ -44,6 +44,8 @@ import org.springframework.dao.InvalidDataAccessApiUsageException;
 import org.springframework.dao.OptimisticLockingFailureException;
 import org.springframework.dao.support.PersistenceExceptionTranslator;
 import org.springframework.data.convert.EntityReader;
+import org.springframework.data.domain.OffsetScrollPosition;
+import org.springframework.data.domain.Scroll;
 import org.springframework.data.geo.Distance;
 import org.springframework.data.geo.GeoResult;
 import org.springframework.data.geo.GeoResults;
@@ -64,6 +66,7 @@ import org.springframework.data.mongodb.core.QueryOperations.DeleteContext;
 import org.springframework.data.mongodb.core.QueryOperations.DistinctQueryContext;
 import org.springframework.data.mongodb.core.QueryOperations.QueryContext;
 import org.springframework.data.mongodb.core.QueryOperations.UpdateContext;
+import org.springframework.data.mongodb.core.ScrollUtils.KeySetCursorQuery;
 import org.springframework.data.mongodb.core.aggregation.Aggregation;
 import org.springframework.data.mongodb.core.aggregation.AggregationOperationContext;
 import org.springframework.data.mongodb.core.aggregation.AggregationOptions;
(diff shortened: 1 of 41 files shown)


christophstrobl:
Queries that either _sort_ by or _project_ on _synthetic_ fields, such as `{ $meta: "textScore" }`, cannot utilize keyset pagination.
Sorting by eg. `{ score: { $meta: "textScore" } }` would need us to include the last seen `score` in the request for the next `Window` leading to something like `{"$text": {"$search": "spring"}, "score" : {"$gt" : 0.5 }}` which returns an empty result cause the stored documents do not contain a `score` field.
To support this feature we'd have to internally convert the `Query` into an `Aggregation` that uses a `[$match, $project, $match]` pipeline that runs the the raw query, additionally projects on the `score` and then matches against the `score` in the third stage. It would be good the get some input from MongoDB if they see value in this approach.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ReactiveMongoTemplate.java
```diff
@@ -826,6 +831,60 @@ public <T> Flux<T> find(@Nullable Query query, Class<T> entityClass, String coll
...
+	public <T> Mono<CursorWindow<T>> findWindow(CursorRequest cursorRequest, Query query, Class<T> entityType) {
+
+		Assert.notNull(entityType, "Entity type must not be null");
+
+		return findWindow(cursorRequest, query, entityType, getCollectionName(entityType));
+	}
+
+	@Override
+	// TODO: CursorRequest and Query declare both a Sort. Which one has precedence?
+	// CursorRequest is similar to Pageable in the sense of being required to define a sort order
```
christophstrobl, line 844: That we have to ask ourselves this is an indicator that maybe the request should not be allowed to redefine sorting.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
```diff
@@ -847,6 +852,62 @@ public <T> List<T> find(Query query, Class<T> entityClass, String collectionName
...
+		}
+
+		if (cursorRequest instanceof KeysetCursorRequest keyset) {
+
+			int limit = keyset.getSize() + 1;
+
+			KeySetCursorQuery keysetPaginationQuery = CursorUtils.createKeysetPaginationQuery(query, keyset,
+					operations.getIdPropertyName(sourceClass));
+
+			List<T> result = doFind(collectionName, createDelegate(query), keysetPaginationQuery.query(),
```
christophstrobl, line 901: we're ignoring the `targetClass` here. Might be worth to call `doFind(createDelegate(query), collectionName, keysetPaginationQuery.query(), keysetPaginationQuery.fields(), sourceClass, targetClass, new QueryCursorPreparer(query, keysetPaginationQuery.sort(), limit, 0, sourceClass));` instead.

christophstrobl:
field projections using expressions work fine since those cannot be used for sorting.
```java
q.fields().project(valueOf("firstName").length()).as("nameLength").include("firstName", "age");
```

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
mp911de: `targetClass` is considered in `ReadDocumentCallback`. We might want to switch to `ProjectingReadCallback` though. Same for the imperative template.

Issue #4312, opened 2023-03-02
Cannot read associations using resolved documents


OddyUlrich:
Good afternoon, I am currently doing a project where I am using an "aggregation" to do a $lookup and a $unwind to get the data from a DBRef with ```lazy=true```. I am doing this as I need the data from one attribute inside another attribute.

```
@Aggregation(pipeline = {
            "{ '$lookup': {" +
                    "'from': 'exerciseBatteries'," +
                    "'localField': 'exerciseBattery.$id'," +
                    "'foreignField': '_id'," +
                    "'as': 'exerciseBattery'} }",
            "{'$unwind': {path: '$exerciseBattery'} }",
            "{'$match': {'exerciseBattery.subject.$id': { $in: ?0} } }"
    })
```

This correctly returns me a list with objects of the class annotated with the ```@Document``` tag and with the attribute marked as a DBRef as an instance with the correct data, but this data is inside the CGLIB$CALLBACK_0 field. This data is fully accessible and there is no problem with it except when I try to store the class with the DBRef in my database.

What I understand should happen is that it should store the object with the reference (DBRef) without problem, however it returns me the following exception: 
```Caused by: org.springframework.data.mapping.MappingException: Cannot create a reference to an object with a NULL id.```

This is because the fields of the class referenced by the DBRef are all set to null, not the ones inside CGLIB$CALLBACK_0, which do not generate any problem, but the fields of the class itself.

![Captura de pantalla 2023-03-02 223954](https://user-images.githubusercontent.com/38378711/222561939-73205a76-f30c-4b45-b69f-b59fecec40e4.png)

Couldn't the data of CGLIB$CALLBACK_0 be copied in those fields to avoid the error or make that when using "save" in a repository it uses, instead, the data of the DGLIB$CALLBACK_0 field?
I would find it quite useful, plus, this operation works perfectly with ```lazy = false``` since no proxy is created and the fields are filled in naturally as with any other class

christophstrobl:
It would be great if you could take the time to provide a complete [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) (something that we can unzip or git clone, build, and deploy) that reproduces the problem.

christophstrobl:
Good catch - thanks for reporting!

OddyUlrich:
Glad to help!

PR #4323, opened 2023-03-09
Allow reading already resolved DBRef values.

christophstrobl:
This PR adds the ability to read (eg. by an aggregation `$lookup` stage) already fully resolved references between documents.
No proxy will be created for lazy loading references and we'll also skip the additional server roundtrip to load the reference by its `_id`.

The change only effects usage of `@DBRef` and does not change handling of `@DocumentReference` which do not have a dedicated type that would allow to distinguish between already resolved and raw references.

Since this is changing the behaviour of how we deal with `DBRef`s I'd keep it on _main_ for the time being and not back port it to any of the maintenance branches.

Closes: #4312

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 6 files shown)


mp911de:

Issue #4080, opened 2022-06-10
Building criteria with in clause against field annotated with `@ValueConverter` causes a ClassCastException


caspianb:
Assume you have an enum that you want to persist as an integer:

```
@ValueConverter(RecordConverter.class)
private RecordEnum record;
...
public static class RecordConverter implements PropertyValueConverter<RecordEnum, Integer, ValueConversionContext<?>> {

    @Override
    public RecordEnum read(Integer value, ValueConversionContext<?> context) {
        return RecordEnum.values()[value];
    }

    @Override
    public Integer write(RecordEnum value, ValueConversionContext<?> context) {
        return value.ordinal();
    }
}
```

If you try to build a query against this field, it will cause cause a ClassCastException:

```
var criteria = Criteria.where("record").in(List.of(RecordEnum.VALUE1, RecordEnum.VALUE2));
mongoTemplate.find(Query.query(criteria), EntityDocument.class);
```

Exact stack trace:
```
java.lang.ClassCastException: class java.util.ImmutableCollections$List12 cannot be cast to class com.logicalbias.domain.RecordEnum (java.util.ImmutableCollections$List12 is in module java.base of loader 'bootstrap'; com.logicalbias.domain.RecordEnum is in unnamed module of loader 'app')

	at com.logicalbias.domain.ParentDocument$RecordConverter.write(ParentDocument.java:63)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedValue(QueryMapper.java:439)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedKeyword(QueryMapper.java:414)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedObjectForField(QueryMapper.java:338)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedObject(QueryMapper.java:167)
	at org.springframework.data.mongodb.core.QueryOperations$QueryContext.getMappedQuery(QueryOperations.java:288)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2660)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2649)
	at org.springframework.data.mongodb.core.MongoTemplate.find(MongoTemplate.java:902)
	at org.springframework.data.mongodb.core.MongoTemplate.find(MongoTemplate.java:888)
[...]

PR #4324, opened 2023-03-09
Fix property value conversion for $in criteria. 

christophstrobl:
This PR fixes an issue where a property `ValueConverter` is not applied if the query is using an `$in` clause that compares the value against a `Collection` of potential candidates that need to run through the converter one by one.

Closes: #4080


mp911de:

Issue #3917, opened 2021-12-21
Sort without grouped fields


1053531172:
db.rep_report_param_label.aggregate([{
            "$match": {}
        }, {
            "$group": {
                "_id": {
                    "_id:null": "$_id:null",
                    "label_name": "$label_name"
                }
            }
        }, {
            "$project": {
                "label_name": "$_id.label_name",
                "_id": 0
            }
        }, {
            "$sort": {
                "serial_number": - 1,
                "label_name": - 1
            }
        }, {
            "$skip": 290
        }, {
            "$limit": 10
        }]) 

this is ok in mongodb.
but i use in spring data mongodb v3.0.0 like this,
 
        GroupOperation groupOperation = Aggregation.group("_id:null", "label_name");
        aggregationOperations.add(groupOperation);
        countAggregationOperations.add(groupOperation);

        ProjectionOperation projectionOperation = Aggregation.project("label_name").andExclude("_id");
        aggregationOperations.add(projectionOperation);
        countAggregationOperations.add(projectionOperation);

        Sort sort = Sort.by(Sort.Direction.DESC, "serial_number");
        SortOperation sortOperation = new SortOperation(sort).and(Sort.Direction.DESC, "label_name");
        aggregationOperations.add(sortOperation);

it was wrong

        java.lang.IllegalArgumentException: Invalid reference 'serial_number'!
	        at org.springframework.data.mongodb.core.aggregation.ExposedFieldsAggregationOperationContext.getReference(ExposedFieldsAggregationOperationContext.java:114) ~[spring-data-mongodb-3.0.0.RELEASE.jar:3.0.0.RELEASE]
	        at org.springframework.data.mongodb.core.aggregation.ExposedFieldsAggregationOperationContext.getReference(ExposedFieldsAggregationOperationContext.java:86) ~[spring-data-mongodb-3.0.0.RELEASE.jar:3.0.0.RELEASE]
	        at org.springframework.data.mongodb.core.aggregation.InheritingExposedFieldsAggregationOperationContext.resolveExposedField(InheritingExposedFieldsAggregationOperationContext.java:62) ~[spring-data-mongodb-3.0.0.RELEASE.jar:3.0.0.RELEASE]
[...]

PR #4328, opened 2023-03-14
Fix field resolution for ExposedFieldsAggregationContext.

christophstrobl:
This PR fixes an issue where the context is not relaxed and errors on unknown fields if multiple stages of nesting contexts happen.

Closes: #3917

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/ExposedFieldsAggregationOperationContext.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/ExposedFieldsAggregationOperationContext.java
@@ -96,6 +96,25 @@ class ExposedFieldsAggregationOperationContext implements AggregationOperationCo
 			return exposedField;
 		}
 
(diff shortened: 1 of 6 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/ExposedFieldsAggregationOperationContext.java
```diff
@@ -96,6 +96,25 @@ private FieldReference getReference(@Nullable Field field, String name) {
 			return exposedField;
 		}
 
+		if(rootContext instanceof RelaxedTypeBasedAggregationOperationContext) {
```
mp911de, line 99: How about `rootContext.getReference(…)`? Wouldn't that work as well?

christophstrobl: IIRC it broke one of the existing tests for `_id` field projection.

mp911de:

Issue #2821, opened 2018-04-09
Add support for reactive bulk operations [DATAMONGO-1922]


spring-projects-issues:
**[Mark Paluch](https://jira.spring.io/secure/ViewProfile.jspa?name=mp911de)** opened **[DATAMONGO-1922](https://jira.spring.io/browse/DATAMONGO-1922?redirect=false)** and commented

---

1 votes, 2 watchers

tgrushka:
Bulk upsert would be wonderful, as `insertAll` fails with a `DuplicateKeyException` if even one record is duplicated, and this exception cannot seem to be caught and ignored, such as in the following code:
```kt
private fun <T: DataModel> insertAllValid(uri: String, transformer: (JSONObject) -> InputModel<T>) {
    getJSON(uri).map(transformer)
        .map(InputModel<T>::toDataModelIfValid)
        .mapNotNull { it!! }
        .collectList()
        // Would be nice to have `::upsertAll` that bulk inserts new + updates existing records
        .flatMapMany(mongoTemplate::insertAll)
        .onErrorResume(DuplicateKeyException::class.java) { e ->
            println(e)
            Flux.empty()
        }
        .doOnNext { logger.debug(it.toString()) }
        .doOnError { logger.error(it.toString()) }
        .subscribe()
}
```

almogtavor:
@tgrushka A nice workaround:
```java
reactiveMongoTemplate.getCollection(config.getCollection()).flatMap(mongoCollection -> {
        var operations = entities.stream().map(entity -> {
            Document doc = new Document();
            reactiveMongoTemplate.getConverter().write(entity, doc);
            var filter = new Document("externalId", entity.getExternalId());
            return new UpdateOneModel<Document>(filter, new Document("$set", doc), new UpdateOptions().upsert(true));
        }).toList();
        return Mono.from(mongoCollection.bulkWrite(operations));
    })
```

argmnt:
@almogtavor Instead of upsert how can I update like this way
var bulkOps = mongoTemplate.bulkOps()
for(dto : List<DTO> DTOs) {
  Query query = new Query();
  query.addCriteria(Criteria.where(ID).is(dto.getId()));
  Update update = new Update()
                    .set(STATUS, dto.getStatus())
  bulkOps.updateOne(query, update)
}
bulkOps.execute();

PR #4342, opened 2023-03-23
Initial support for reactive bulk operations.

christophstrobl:
Closes: #2821

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/DefaultReactiveBulkOperations.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/DefaultReactiveBulkOperations.java
@@ -1,0 +1,591 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/DefaultReactiveBulkOperations.java
```diff
@@ -0,0 +1,591 @@
...
+							List<Mono<Object>> monos = it.stream().map(saved -> {
+								return maybeInvokeAfterSaveCallback(saved);
+							}).collect(Collectors.toList());
+
+							return Flux.concat(monos).then(Mono.just(state));
+						});
+			});
+		} catch (RuntimeException ex) {
+
+			if (ex instanceof MongoBulkWriteException) {
```
mp911de, line 257: Can these even happen? Why not `catch(MongoBulkWriteException e)`?

christophstrobl: if the operation is `ordered`, yes.
https://www.mongodb.com/docs/manual/reference/method/db.collection.bulkWrite/#error-handling

mp911de: But in reactive, errors are emitted as signals through the publisher (that's why I was wondering) and not through the synchronous calls and therefore, the error signal translation is missing.

christophstrobl: doh - you're right

mp911de: Okay, thanks for confirming. I'm happy to handle that aspect during the merge.

Issue #4351, opened 2023-03-29
Issue with AOT and Kotlin data class: The following must be non-sealed interfaces


fab918:
When I add the gradle AOT plugin `id("org.graalvm.buildtools.native")` I have an error on a simple `data class` document from `org.springframework.data.mongodb.aot.LazyLoadingProxyAotProcessor`:

```java
Exception in thread "main" java.lang.IllegalArgumentException: The following must be non-sealed interfaces: [...persistence.entity.Tags]
	at org.springframework.aot.hint.JdkProxyHint$Builder.toTypeReferences(JdkProxyHint.java:159)
	at org.springframework.aot.hint.JdkProxyHint$Builder.proxiedInterfaces(JdkProxyHint.java:129)
	at org.springframework.aot.hint.ProxyHints.lambda$registerJdkProxy$1(ProxyHints.java:78)
	at org.springframework.aot.hint.ProxyHints.registerJdkProxy(ProxyHints.java:50)
	at org.springframework.aot.hint.ProxyHints.registerJdkProxy(ProxyHints.java:77)
	at org.springframework.data.mongodb.aot.LazyLoadingProxyAotProcessor.lambda$registerLazyLoadingProxyIfNeeded$2(LazyLoadingProxyAotProcessor.java:76)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1845)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
[...]

PR #4352, opened 2023-03-31
Fix AOT processing for lazy loading jdk proxies.

christophstrobl:
Closes: #4351 
Integration test: spring-projects/spring-aot-smoke-tests#176

Should also be forward ported to _main_ development line.


mp911de:

PR #4382, opened 2023-05-02
Fix Querydsl conversion for properties with explicit target type. 

### LINKED ISSUE #4359: QueryDSL-style queries don't honor targetType @Field attribute
I have a `@Document` class, `Foo` with a field like this:
```
@Field(value = "bar", targetType = FieldType.OBJECT_ID)
String bar;
```
I'm using QueryDSL so I also end up with a `QFoo` generated class. In another class, I have a field `r` whose type is a `@Repository` interface `R` that implements `QuerydslPredicateExecutor`. 

When I call `r.findAll(QFoo.foo.bar.eq("...")` I expect the "..." to be converted to an `ObjectId` as part of the construction of the MongoDB query but instead it gets passed through as a string--i.e., the query looks like:
```
{bar: "..."}
```
so the query doesn't work as expected.

christophstrobl:
This PR makes sure to convert values, for _Querydsl_ predicates pointing to a property with an explicitly annotated target type, into the desired format.

Resolves: #4359

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SpringDataMongodbSerializer.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SpringDataMongodbSerializer.java
@@ -27,6 +27,7 @@ import org.springframework.data.mongodb.core.convert.MongoConverter;
 import org.springframework.data.mongodb.core.convert.QueryMapper;
 import org.springframework.data.mongodb.core.mapping.MongoPersistentEntity;
 import org.springframework.data.mongodb.core.mapping.MongoPersistentProperty;
+import org.springframework.data.util.TypeInformation;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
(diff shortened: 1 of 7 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SpringDataMongodbSerializer.java
```diff
@@ -149,12 +150,15 @@ protected String asDBKey(@Nullable Operation<?> expr, int index) {
 
 	protected Object convert(@Nullable Path<?> path, @Nullable Constant<?> constant) {
 
+		MongoPersistentProperty property = getPropertyFor(path);
+
 		if (!isReference(path)) {
+			if(property != null && property.hasExplicitWriteTarget()) {
```
mp911de, line 156: I wonder why this is necessary as the Querydsl predicate is rendered into our `Query` object that is then converted using the domain type information.

christophstrobl: true, the query mapper will take care of the conversion. So it works fine on current _main_, let me get back to the reporter to see if that's an issue with earlier versions. I would have argued that `SpringDataMongodbSerializer` should already consider this based on its description, so it's worth updating that.

PR #4402, opened 2023-06-07
Add support for MongoDB Kotlin sync/coroutine Driver.

### LINKED ISSUE #4393: Add support for sync MongoDB Kotlin driver
Allow database factories to be created using a Kotlin MongoDB Driver.

```xml
<dependency>
    <groupId>org.mongodb</groupId>
    <artifactId>mongodb-driver-kotlin-sync</artifactId>
    <scope>optional</scope>
</dependency>
```

christophstrobl:
Provide an extension for `SimpleMongoClientDatabaseFactory` & `SimpleReactiveMongoDatabaseFactory` that accepts a `com.mongodb.kotlin.client.MongoClient` respectively `com.mongodb.kotlin.client.coroutine.MongoClient`

Since the Kotlin driver is wrapping the Java driver we can unwrap the instance and pass it on to the factory.

Though unfortunate that there's no public API to access the wrapped instance the provided tests will alert on driver changes. 

Closes: #4393

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.2.0-SNAPSHOT</version>
+	<version>4.2.x-4393-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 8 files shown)


christophstrobl:
@rozza would it be possible to have some public API in the Kotlin driver to access the underlying Java driver? Currently we need to use reflection to get hold on the `wrapped` client.

PR #4406, opened 2023-06-12
Map collection and fields for `$lookup`/`$graphLookup` aggregation stage against domain type

### LINKED ISSUE #4379: Expose method that calculates collection name from class
Working on a legacy code base, we've got lots of code that builds up aggregations including lookups, so we want to lookup the collection name for a class.

We've got an implementation that I'm guessing was at some point copied from mongo, but doesn't seem up to date and digging through code can assume it was copied from 
**BasicMongoPersistentEntity**

```
	public BasicMongoPersistentEntity(TypeInformation<T> typeInformation) {

		super(typeInformation, MongoPersistentPropertyComparator.INSTANCE);

		Class<?> rawType = typeInformation.getType();
		String fallback = MongoCollectionUtils.getPreferredCollectionName(rawType);

		if (this.isAnnotationPresent(Document.class)) {
			Document document = this.getRequiredAnnotation(Document.class);

			this.collection = StringUtils.hasText(document.collection()) ? document.collection() : fallback;
			this.language = StringUtils.hasText(document.language()) ? document.language() : "";
			this.expression = detectExpression(document.collection());
			this.collation = document.collation();
			this.collationExpression = detectExpression(document.collation());
		} else {

			this.collection = fallback;
			this.language = "";
			this.expression = null;
			this.collation = null;
			this.collationExpression = null;
		}
```

Is there any reason this isn't exposed as a static method call so anyone can resolve class name to collection name consistently with how mongo does it?

christophstrobl:
This commit enables using a _type_ parameter to define the from collection of a `$lookup`/`$graphLookup` aggregation stage. In doing so we can derive the target collection name from the type and use the given information to also map the _from_ field against the domain object to so that the user is able to operate on property names instead of the target document field name.

Resolves: #4379

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
@@ -23,6 +23,7 @@ import org.bson.Document;
 import org.bson.codecs.configuration.CodecRegistry;
 import org.springframework.beans.BeanUtils;
 import org.springframework.data.mongodb.CodecRegistryProvider;
+import org.springframework.data.mongodb.MongoCollectionUtils;
 import org.springframework.data.mongodb.core.aggregation.ExposedFields.FieldReference;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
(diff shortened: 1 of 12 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationContext.java
```diff
@@ -78,6 +79,29 @@ default Document getMappedObject(Document document) {
...
 
+	/**
+	 * Obtain the target field name for a given field/type combination.
+	 * 
+	 * @param type The type containing the field.
+	 * @param field The property/field name
+	 * @return never {@literal null}.
+	 * @since 4.2
+	 */
+	default String getMappedFieldName(Class<?> type, String field) {
```
mp911de, line 90: Instead of adding another method that accepts a type, how about either `AggregationOperationContext.withType(…)` or reusing `getFields(Class<?> type)`?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/TypeBasedAggregationOperationContext.java
```diff
@@ -92,6 +92,20 @@ public FieldReference getReference(String name) {
...
+	public String getCollection(Class<?> type) {
+
+		MongoPersistentEntity<?> persistentEntity = mappingContext.getPersistentEntity(type);
+		return persistentEntity != null ? persistentEntity.getCollection() : AggregationOperationContext.super.getCollection(type);
+	}
+
+	@Override
+	public String getMappedFieldName(Class<?> type, String field) {
+
+		PersistentPropertyPath<MongoPersistentProperty> persistentPropertyPath = mappingContext.getPersistentPropertyPath(field, type);
```
mp911de, line 105: `RelaxedTypeBasedAggregationOperationContext` doesn't consider any relaxed handling. Ideally, we could find an approach that embeds into our mandatory vs. relaxed field handling.

Issue #4404, opened 2023-06-08
`InIgnoreCase` queries match on substrings


zlaval:
The generated query is not ok in case of using `in` expression with `IgnoreCase`.

Example:

```
@Document(collection = "mydata")
data class MyData(
   @Id
    val id: String,
    val name: String,
)
```

I have a list of names and I would like to query for documents which name is in the list case insensitively.

Using `fun findAllByNameInIgnoreCase(names: List<String>)` creates the following mongo query:

```
"name" : { "$in" : [{ "$regularExpression" : { "pattern" : "\\QTest One\\E", "options" : "i"}}, { "$regularExpression" : { "pattern" : "\\QTest Two\\E", "options" : "i"}}]}
```

The problem it will also match to `Test Two Something`.

This works just fine in case of `fun findByNameIgnoreCase(names: String)`, which generates this query:

```
"name" : { "$regularExpression" : { "pattern" : "^Test One$", "options" : "i"}}
```

So i think this would be consistent if the generated query was:

```
"name" : { $in : [{ "$regularExpression" : { "pattern" : "^\\QTest One\\E$", "options" : "i"}}, { "$regularExpression" : { "pattern" : "^\\QTest Two\\E$", "options" : "i"}}]}
```

This would results in the expected result, as the query method does not contains `Regex` expression even if mongo solves case insensivity with regex.

christophstrobl:
Thanks @zlaval for raising the issue.

PR #4412, opened 2023-06-13
Use exact matching for `IN` clause with ignore case

christophstrobl:
Prior to this change the generated pattern would have matched more entries than it should have. The behavior is now aligned to its counterpart not using the `IgnoreCase` flag.

Resolves: #4404


mp911de:

Issue #4429, opened 2023-06-27
Refine locking behaviour of core components.


christophstrobl:
* MessageListenerContainer
* Cursors
* LazyLoading

PR #4431, opened 2023-06-27
Refine Locking

christophstrobl:
Closes: #4429

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/LazyLoadingProxyFactory.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/LazyLoadingProxyFactory.java
@@ -22,6 +22,10 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.lang.reflect.Method;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Supplier;
 
 import org.aopalliance.intercept.MethodInterceptor;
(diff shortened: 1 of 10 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/LazyLoadingProxyFactory.java
```diff
@@ -349,13 +356,15 @@ private synchronized Object resolve() {
 				}
 				return result;
 			}
+			lock.readLock().unlock();
```
jxblum, line 359: The "read" lock should be released inside a `finally { .. }` block here, particularly since there are interactions with a `Logger` occurring just above it.

If the `LOGGER` was not properly initialized, or if the `property` reference were `null` when trace is enabled, or some other issues with logging occurred and threw an Exception, then the "read" lock would not be properly released, and it would prevent any writes from occurring thereby leading to possible Thread starvation.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/messaging/CursorReadingTask.java
```diff
@@ -39,7 +41,7 @@
  */
 abstract class CursorReadingTask<T, R> implements Task {
 
-	private final Object lifecycleMonitor = new Object();
```
mp911de, line 42: `synchronized` blocks within this class serve the purpose to ensure ordering according (happens before as of the JLS spec). We should keep these to avoid visibility issues of the `state` field.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/messaging/DefaultMessageListenerContainer.java
```diff
@@ -171,35 +180,42 @@ public <S, T> Subscription register(SubscriptionRequest<S, ? super T, ? extends
 	@Override
 	public Optional<Subscription> lookup(SubscriptionRequest<?, ?, ?> request) {
 
-		synchronized (lifecycleMonitor) {
+		subscriptionMonitor.readLock();
```
jxblum, line 183: Should be `subscriptionMonitor.readLock().lock();`

christophstrobl:
rebased & updated. I would not want to introduce more public API but moved some code into methods with limited scope.

Issue #4454, opened 2023-07-18
Schema creator may miss encryption on entity field.


christophstrobl:
The schema creator may miss to wrap `JsonSchemaProperty` into `EncryptedJsonSchemaProperty` for properties that are considered entities. 

The following input
```java
@Encrypted(algorithm = ...)
class X {
    // ...
    @Encrypted Y yValue;
}
class Y {
    String value;
}
```
should create a schema as follows.

```json
{
    "encryptMetadata" : { "algorithm" : "..." }, 
    "type" : "object", 
    "properties" : {
        "yValue" : { "encrypt" : { "bsonType": "object" } }
    }
}
```

PR #4455, opened 2023-07-18
Fix schema generation for encrypted fields that are considered domain entities.

christophstrobl:
This PR makes sure to consider the encrypted annotation on fields that are considered domain type property values, encrypting the entire object if necessary.

Closes: #4454


sxhinzvc:
Merged and backported. Thank you!

PR #4486, opened 2023-08-28
Provide `DocumentReferenceSource` to document reference lookup for use of properties in constructor creation

sxhinzvc:
This commit enables a source entity with a one-to-many @DocumentReference lookup property to be instantiated properly when the referenced entity - which contains the linking value - is used as a constructor argument.


mp911de:
I updated the ticket title. One-to-many didn't fit necessarily as we can have single-valued references as well.

Issue #2067, opened 2015-01-28
Introduce ability to append queries against a particular collection [DATAMONGO-1151]


spring-projects-issues:
**[Heiko Scherrer](https://jira.spring.io/secure/ViewProfile.jspa?name=scherrer)** opened **[DATAMONGO-1151](https://jira.spring.io/browse/DATAMONGO-1151?redirect=false)** and commented

**Task description**
I request a feature that I couldn't find in the current implementation. Probably there's another elegant way to solve the requested behaviour. I need to append all queries to a particular collection with a specific criteria. Think about a multi-tenancy scenario where a discriminator property (holding the tenants name) has to be compared in each query. Doing this within application code is cumbersome and error-prone. IMO this should be applied somewhere deep in the framework, at best near to execution.

**Investigation**
I had a look at MongoTemplate and implemented a solution with a so called QueryAppender, that is able to add a criteria to a query (for a collection). Unfortunately MongoTemplate uses several alternatives within the finder methods. Some work with Query, others with DBObjects. So before opening Pandoras Box and implementing too many changes within MongoTemplate, I wanted to clarify that my approach might be a solution and no other solution already exists.

---

**Affects:** 1.6.2 (Evans SR2)

**Issue Links:**
- [DATACMNS-293](https://jira.spring.io/browse/DATACMNS-293) Add infrastructure for generic query augmentation
 (_**"depends on"**_)
- [DATAJPA-307](https://jira.spring.io/browse/DATAJPA-307) Add support for soft deletes

- [DATAJPA-564](https://jira.spring.io/browse/DATAJPA-564) Support for SpEL based parameter expressions in repository query methods

ArpanKIIT2017:
Is there any updates on this? Even I need the tenantId based multi-tenancy support. We figured it is too confusing to find a place to append the extra criteria.

Just wanted to understand, does implementing a mongo template also automatically scope the autogenerated queries from arbitrary interface method names?

christophstrobl:
@ArpanKIIT2017 Every repository query will be run by the template, though there's no easy way to work with the `QueryMapper` that is used. Nor is there support for augmenting the query, which is as you already know an open issue in data-commons.

PR #4498, opened 2023-09-12
MongoTemplate doFind for ad-hoc query method visibility as protected.

aleixmorgadas:
Need:
The method that's being called when using method-like queries has default package visibility, compared to other methods with protected visibility.

In order to extend `MongoTemplate` without placing the new class in the `org.springframework.data.mongodb.core` package, I added the `protected` as method visibility.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
@@ -2606,7 +2606,7 @@ public class MongoTemplate
 	 *
 	 * @since 2.0
 	 */
-	<S, T> List<T> doFind(CollectionPreparer<MongoCollection<Document>> collectionPreparer, String collectionName,
+	protected <S, T> List<T> doFind(CollectionPreparer<MongoCollection<Document>> collectionPreparer, String collectionName,
 			Document query, Document fields, Class<S> sourceClass, Class<T> targetClass, CursorPreparer preparer) {
 
 		MongoPersistentEntity<?> entity = mappingContext.getPersistentEntity(sourceClass);


mp911de:
This is by design as we do not want to expose the method. It has a very large number of arguments and exposing it removes our desire to freely change the method.

Let's take a step back: Can you elaborate on what you're trying to achieve?

aleixmorgadas:
Sure!

We had to do some work around add filters to the query due to our system being multitenant and we wanted to avoid to add `ifs` everywhere to verify that the fetched data belongs to a tenant.

We extended `MongoTemplate` and added some annotations to allow that behaviour. The result is this OSS library.

https://github.com/aleixmorgadas/spring-boot-data-mongo-multitenant

You can see that we added the [specific class](https://github.com/aleixmorgadas/spring-boot-data-mongo-multitenant/blob/main/src/main/java/org/springframework/data/mongodb/core/MultiTenantMongoTemplate.java) into `org.springframework.data.mongodb.core` in order to extend the method as a work around.

The library isn't the best but it works.

christophstrobl:
@aleixmorgadas thanks for providing additional information - In the light of this PR I think it would make sense to revisit  #2067 and maybe even spring-projects/spring-data-commons#766.

aleixmorgadas:
I'm happy to help here.

Right now, I took the approach of extending the behaviour via the `@MultiTenant` annotation at the Entity level. I'm considering do it at the Repository level instead.

I think it helps making the Repository more obvious by being:
```java
interface UserRepository extends MultiTenantCrudRepository<User, ObjectId> {
    String tenantField() {
        return "tenant";
    }
}
```

Instead of
```java
interface UserRepository extends CrudRepository<User, ObjectId> {
}

@MultiTenant
@Document
class User {
    @Id 
    ObjectId id;

    @MultiTenantField
    String tenant;

oscarbrookssynacor:
Hi there, 
This very simple change is currently blocking multiple long standing, and I would argue, highly impactful issues, some of which have been opened and asked about for over 10 years. 
https://github.com/spring-projects/spring-data-commons/issues/766
https://github.com/spring-projects/spring-security/issues/2629
https://github.com/spring-projects/spring-security/issues/6736

All of these ultimately are blocked by this change. In light of that I would like to request that this be given a higher priority.

mp911de:
There's some tension in exposing internal methods because we have over time a lot of changes to internals. If your application would rely on these methods, these would break quite often. Instead it would make rather sense to introduce query post-processing hooks with a well-defined API.

oscarbrookssynacor:
Hi, thanks for the quick reply, I understand your position, my goal is to simply try and facilitate getting the ball moving on those tickets I listed, I am not partial to the way this tickets gets resolved. 
The option you are suggesting, what would it take to get that accomplished? I am fairly well versed in spring internals, but I have not worked with data-mongodb before.

oscarbrookssynacor:
To add, it seems like there are no hard blocker to accomplishing any of the features on any of those tickets, and it seems like there is simply a lack of cross-communication between the projects, perhaps it would help to get all the respective folks involved in one conversation.

Issue #4510, opened 2023-09-21
Query defect in spring-data-mongodb QueryMapper with annotations and PropertyValueConverter


judby:
- JAVA 17
- spring-boot 3.0.11
- spring-data-mongodb 4.0.10

Introduced a `@LocalDateAsString` annotation used on selected `LocalDate` fields to have them mapped to string in mongo.

And a `LocalDateToStringPropertyConverter`which simply converts to/from String/LocalDate.

Everything is fine with mapping and simple queries but with eg "range" queries the `QueryMapper` calls the converter with a bson `Document` which obviously fails with a `ClassCastException`.

The annotation/converter implementations are trivial, [code can be found on GitHub](https://github.com/judby/defect-spring-data-mongodb-query/blob/main/src/main/java/com/udby/defect/spring/data/mongodb/converter/query/simple/).

When a query like below is executed targeting a field that has aforementioned annotation:

```
final var query = Query.query(where("date").gte(LocalDate.parse("2023-09-20")).lte(LocalDate.parse("2023-09-22")));

final var testEntities = mongoTemplate.find(query, TestEntity.class);
```

an exception is thrown in `QueryMapper` since it tries to invoke the converter with the bson `Document` representing the `gte` and `lte` expressions:

```
java.lang.ClassCastException: class org.bson.Document cannot be cast to class java.time.LocalDate (org.bson.Document is in unnamed module of loader 'app'; java.time.LocalDate is in module java.base of loader 'bootstrap')

	at com.udby.defect.spring.data.mongodb.converter.query.simple.LocalDateToStringPropertyConverter.write(LocalDateToStringPropertyConverter.java:10)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedValue(QueryMapper.java:450)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedObjectForField(QueryMapper.java:339)
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedObject(QueryMapper.java:166)
	at org.springframework.data.mongodb.core.QueryOperations$QueryContext.getMappedQuery(QueryOperations.java:350)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2422)
	at org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2411)
	at org.springframework.data.mongodb.core.MongoTemplate.find(MongoTemplate.java:815)
[...]

PR #4517, opened 2023-09-28
Fix property value conversion in query mapper for nested values.

christophstrobl:
Resolves: #4510

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/QueryMapper.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/QueryMapper.java
@@ -86,6 +86,7 @@ import com.mongodb.DBRef;
  * @author Mark Paluch
  * @author David Julia
  * @author Divya Srivastava
+ * @author Gyungrai Wang
  */
 public class QueryMapper {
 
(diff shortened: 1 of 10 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/QueryMapper.java
```diff
@@ -455,6 +455,15 @@ public <T> T getPropertyValue(MongoPersistentProperty property) {
 				return collection.stream().map(it -> valueConverter.write(it, conversionContext)).collect(Collectors.toList());
 			}
 
+			if(!documentField.getProperty().isMap() && sourceValue instanceof Document document) {
+				return new Document(document.entrySet().stream().collect(Collectors.toMap(entry -> entry.getKey(), entry -> {
```
judby, line 459: minor suggestion: `.collect(Collectors.toMap(Entry::getKey, entry -> {`

christophstrobl: We had a second look and using the stream API here bears some risks. Entry order may not be preserved and the usage of Collectors API can cause NPE due to [JDK-8148463](https://bugs.openjdk.org/browse/JDK-8148463).

Issue #4524, opened 2023-10-09
Repository query not working since 3.4.x


jcedenopyx:
When using a `MongoRepository<A, B>`, query functions that project from type A to type C (with compatible fields), spring attempts to convert the full object to the Id field type. 

This behavior worked fine until spring data 3.3.x. Using `@Id` or `@MongoId` on the Id field has no effect, but using `@Field("_id")` makes it so that it maps properly.

Demo reproducing the bug: https://github.com/jcedenopyx/spring-data-mongo-query-projection-issue/blob/master/src/test/java/me/jcedeno/demo/DemoApplicationTests.java#L89-L95

```
org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [me.jcedeno.demo.models.Transaction] to type [me.jcedeno.demo.models.TransactionId]
	at app//org.springframework.core.convert.support.GenericConversionService.handleConverterNotFound(GenericConversionService.java:322)
	at app//org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:195)
	at app//org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:175)
	at app//org.springframework.data.mapping.model.ConvertingPropertyAccessor.convertIfNecessary(ConvertingPropertyAccessor.java:121)
	at app//org.springframework.data.mapping.model.ConvertingPropertyAccessor.setProperty(ConvertingPropertyAccessor.java:64)
	at app//org.springframework.data.mongodb.core.convert.MappingMongoConverter.readAndPopulateIdentifier(MappingMongoConverter.java:574)
	at app//org.springframework.data.mongodb.core.convert.MappingMongoConverter.populateProperties(MappingMongoConverter.java:541)
	at app//org.springframework.data.mongodb.core.convert.MappingMongoConverter.doReadProjection(MappingMongoConverter.java:377)
	at app//org.springframework.data.mongodb.core.convert.MappingMongoConverter.project(MappingMongoConverter.java:323)
	at app//org.springframework.data.mongodb.core.MongoTemplate$ProjectingReadCallback.doWith(MongoTemplate.java:3348)
	at app//org.springframework.data.mongodb.core.MongoTemplate.executeFindMultiInternal(MongoTemplate.java:2959)
	at app//org.springframework.data.mongodb.core.MongoTemplate.doFind(MongoTemplate.java:2676)
[...]

PR #4525, opened 2023-10-11
Fix id value conversion when projecting result types.

christophstrobl:
Contextual information required for converting values are now passed on correctly when projecting `id` properties.

Resolves: #4524


mp911de:

jcedenopyx:
Is this going out on October 13th?

Issue #4495, opened 2023-09-06
`ReactiveMongoTemplate` does not return `ChangeStreamEvents` with PreImages


PhilippAllstadt:
Hi @all,

in our use case we have the requirement to consume changestream events with pre-images. We are using MongoDB version 6.0.7.
Further, our spring version is the newest and we are using the latest mongodb drivers. 

Despite it, delete and update events haven´t filled fullDocumentBeforeChange in raw if event is consumed in our spring app.

In database. pre-images should work because we´ve tested consuming via mongosh shell. 

This is our code to consume change stream in our spring app:

 `var collectionChangeStream =
        reactiveMongoTemplate
            .changeStream(TransactionEntity.class)
            .withOptions(
                (optionsBuilder) ->
                    optionsBuilder
                        .returnFullDocumentBeforeChange()
                        .returnFullDocumentOnUpdate())
            .watchCollection("transaction");`

    if (changeStreamToken.isPresent() && changeStreamToken.get().getToken() != null) {
      collectionChangeStream.resumeAfter(changeStreamToken.get().getToken());
    } else {
      collectionChangeStream.resumeAt(Instant.now().minusSeconds(seconds));
    }

    return collectionChangeStream.listen();

Am I missing something?

Thank you very much.

christophstrobl:
Thank you @PhilippAllstadt for bringing this to our attention. 
At a quick glance it seems `ReactiveMongoTemplate#changeStream(String, String, ChangeStreamOptions, Class)` does not forward the `FullDocumentBeforeChange` option to the `ChangeStreamPublisher`.

PhilippAllstadt:
Thank you for your response.

Yes, i think the problem is prepareFilter method:

`public <T> Flux<ChangeStreamEvent<T>> changeStream(@Nullable String database, @Nullable String collectionName,
			ChangeStreamOptions options, Class<T> targetType) {

		List<Document> filter = prepareFilter(options);
		FullDocument fullDocument = ClassUtils.isAssignable(Document.class, targetType) ? FullDocument.DEFAULT
				: FullDocument.UPDATE_LOOKUP;

		return ReactiveMongoDatabaseUtils.getDatabase(database, mongoDatabaseFactory) //
				.map(db -> {
					ChangeStreamPublisher<Document> publisher;
					if (StringUtils.hasText(collectionName)) {
						publisher = filter.isEmpty() ? db.getCollection(collectionName).watch(Document.class)
[...]

PR #4541, opened 2023-11-02
Pass on FullDocumentBeforeChange option to change stream iterable/publisher.

christophstrobl:
This PR ensures to pass on a potentially set `FullDocumentBeforeChange` option to the change stream iterable/publisher.
It also corrects false optional behaviour within the `ChangeStreamTask` which did some defaulting though the actual value is an optional one that must not be present.

Closes: #4495


mp911de:

Issue #4546, opened 2023-11-03
NoSuchElementException throws while using @Aggregation to group 


human-user:
my case :
```
@Aggregation(value = "[{ \"$group\" : { \"_id\" : \"$taskStatus\", \"count\" : { \"$sum\" : 1}}}]")
AggregationResults<CountScanTaskRespDto> countByTaskStatus();
```

It work well while using `mongoTemplate.aggregate` :
```
AggregationResults<CountScanTaskRespDto> result = mongoTemplate.aggregate(
        aggregation, ScanTask.class, CountScanTaskRespDto.class);
return result.getMappedResults();
```

exception stack log: 
```
java.util.NoSuchElementException: null
	at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:758) ~[na:na]
	at java.base/java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:778) ~[na:na]
	at org.springframework.data.mongodb.core.aggregation.AggregationOperation.getOperator(AggregationOperation.java:66) ~[spring-data-mongodb-4.1.1.jar:4.1.1]
	at org.springframework.data.mongodb.core.aggregation.AggregationPipeline.isOut(AggregationPipeline.java:165) ~[spring-data-mongodb-4.1.1.jar:4.1.1]
	at org.springframework.data.mongodb.core.aggregation.AggregationPipeline.isOutOrMerge(AggregationPipeline.java:99) ~[spring-data-mongodb-4.1.1.jar:4.1.1]
	at org.springframework.data.mongodb.repository.query.StringBasedAggregation.computeOptions(StringBasedAggregation.java:185) ~[spring-data-mongodb-4.1.1.jar:4.1.1]
	at org.springframework.data.mongodb.repository.query.StringBasedAggregation.doExecute(StringBasedAggregation.java:112) ~[spring-data-mongodb-4.1.1.jar:4.1.1]
	at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.execute(AbstractMongoQuery.java:118) ~[spring-data-mongodb-4.1.1.jar:4.1.1]
	at org.springframework.data.repository.core.support.RepositoryMethodInvoker.doInvoke(RepositoryMethodInvoker.java:136) ~[spring-data-commons-3.1.1.jar:3.1.1]
[...]

PR #4547, opened 2023-11-03
Validate String based aggregation on query method creation.

christophstrobl:
Related to: #4546

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/MongoQueryMethod.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/MongoQueryMethod.java
@@ -504,6 +504,16 @@ public class MongoQueryMethod extends QueryMethod {
 				}
 			}
 		}
+		if (hasAnnotatedAggregation()) {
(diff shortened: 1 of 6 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/MongoQueryMethod.java
```diff
@@ -504,6 +504,16 @@ public void verify() {
 				}
 			}
 		}
+		if (hasAnnotatedAggregation()) {
+			for (String stage : getAnnotatedAggregation()) {
```
mp911de, line 508: How about using `ParameterBindingJsonReader` to verify that each declaration contains a single object?

christophstrobl: The `QueryMethod` itself does at this point not know about any parameters nor the parser to use so any validation attempt would be quite error prone due to the potential, non resolvable placeholders contained in the source string. 
However there's `BsonUtils#isJsonArray` which we might want to reuse.

PR #4550, opened 2023-11-06
Switch to `Flux.flatMapSequential(…)` to prevent backpressure shaping

mp911de:
We now use `Flux.flatMapSequential(…)` instead of `concatMap` as `concatMap` reduces the request size to `1`. The change in backpressure/request size reduces parallelism and impacts the batch size by fetching `2` documents instead of considering the actual backpressure.

flatMapSequential doesn't tamper the requested amount while retaining the sequence order.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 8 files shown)


christophstrobl:
Merged to _main_ and back ported to _4.0.x_ and _4.1.x_.

PR #4552, opened 2023-11-11
Declarative way for setting MongoDB transaction options

### LINKED ISSUE #1628: Declarative way for setting MongoDB transaction options
**[Björn Voß](https://jira.spring.io/secure/ViewProfile.jspa?name=bjoern.voss)** opened **[DATAMONGO-697](https://jira.spring.io/browse/DATAMONGO-697?redirect=false)** and commented

Hi,

I would like to discuss a feature, maybe called "declarative MongoTx", which I need much more than the possibility of calculating a WriteConcern based on a MongoAction, as I think more in usecases greater than on mongo call.

So mongo is eventually consistent and as a user of mongo you know that and are happy with that, but sometimes you might have the need to do some stuff really against the actual data (primary) and as you do critical stuff you want to be 'REPLICA_SAFE' as the other times reading from secondaries is absolute fine as well as the writeConcern NONE.

so what I have done for our project [yum-repo-server](https://github.com/ImmobilienScout24/yum-repo-server)

```
@MongoTx(readPreference = "primary", writeConcern = "REPLICAS_SAFE")
public void doMyMissionCriticalStuff() {
...
}
```

And now I have a similar behavior like `@Transactional` with spring-data, regarding the isolation level, but NO rollback or so.

What do you think about my idea and do you have similar needs?

regards

---

**Attachments:**
- [MongoTx1.patch](https://jira.spring.io/secure/attachment/21240/MongoTx1.patch) (_21.33 kB_)

ykardziyaka:
Closes #1628

After I read the issue description and thought about it a bit, I figured there's conceptually three possible approaches to implement requested functionality:

1. By adding new attributes to `@Transactional` annotations like `@Transactional(readPreference = PRIMARY)`
2. By introducing new MongoDb-specific annotation for managing transactions 
3. By re-using existing attributes of `@Transactional`

First option requires to add database-specific attributes to database-generic annotation. That doesn't sound right.

Second option can be implemented either by completely re-implementing/duplicating whole `@Transactional` functionality in Spring Data MongoDb or by altering `@Transactional` interceptor so that it can pass attributes of new annotation along with existing values. Even if that's possible, it will be too much effort. Also, splitting transaction configuration into two annotations kinda confusing.

So I implemented third option. The only suitable attribute in `@Transactional` to store MongoDb options is `label`. Spring docs suggest that:

>  Labels may serve a solely descriptive purpose or **map to pre-defined transaction manager-specific options**.  

This PR adds possibility to add 4 labels that will be recognized as transaction options on per-`@Transactional` basis:
- `mongo:maxCommitTime={value}`
- `mongo:readConcern={value}`
- `mongo:readPreference={value}`
- `mongo:writeConcern={value}`

`mongo:readConcern`, `mongo:readPreference`, `mongo:writeConcern` values are converted from String to respective Mongo Client types using Mongo Client API methods.

`mongo:maxCommitTime` is converted from String using `Duration.parse(String)`.

Example of usage:
```java
  @Transactional(transactionManager = "txManager", label = { 
    "mongo:maxCommitTime=PT1M", 
    "mongo:readConcern=majority" ,
    "mongo:readPreference=primary" ,
    "mongo:writeConcern=acknowledged" ,
  })
  public Entity doSomeStuff(Entity entity) {
      //...
  }
```

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoTransactionManager.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoTransactionManager.java
@@ -134,7 +134,7 @@ public class MongoTransactionManager extends AbstractPlatformTransactionManager
 		}
 
 		try {
-			mongoTransactionObject.startTransaction(options);
+			mongoTransactionObject.startTransaction(MongoTransactionUtils.extractOptions(definition, options));
 		} catch (MongoException ex) {
 			throw new TransactionSystemException(String.format("Could not start Mongo transaction for session %s.",
 					debugString(mongoTransactionObject.getSession())), ex);
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoTransactionUtils.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/MongoTransactionUtils.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.mongodb;
+
+import java.time.Duration;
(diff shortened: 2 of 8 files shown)


ykardziyaka:
Missed some formatting in tests, should be fine now

christophstrobl:
Thank you very much @ykardziyaka for taking the time to dig into this topic. I like the approach using the `label` attribute 👍
Since we're already pretty close to the 4.2.0 GA release, we intend to pick up this PR for the next development iteration.

christophstrobl:
Currently planned to be picked up for Milestone 2.
We're also thinking of adding means to customize the the label prefix so that one can move off the default `mongo:`

christophstrobl:
@ykardziyaka you may want to have a look at the current [issue branch](https://github.com/spring-projects/spring-data-mongodb/tree/issue/1628).

mp911de:

Issue #4571, opened 2023-11-28
Behavior breaking change deserializing property with null value


pierre-gautier:
Using spring-data-mongodb 4.1.5, when deserializaing a property with null value in database to an instanciated collection, the collection was not set to null.

With version 4.2.0, the collection is set to null.

While I do understand the intention and tends to be aligned with it, it's still a breaking change that may have numerous impacts in production.

WDYT ?

christophstrobl:
Thank you for reporting! Do you by any chance have a complete [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) (something that we can unzip or git clone, build, and deploy) that reproduces the problem? That would help us quite a bit triaging the issue.

pierre-gautier:
I don't but I'll spend some time on it

pierre-gautier:
Would it be enough if I told you that the difference happens in MappingMongoConverter#readProperties ?
In 4.1.5 on line 610 `documentAccessor.hasValue(prop)` returns false while it returns true in 4.2.0.
Digging down it seems the issue comes from BsonUtils#hasValue that has changed [here](https://github.com/spring-projects/spring-data-mongodb/commit/691fc055edbfb3decc61d2d5ce8ec8ece7753a49#diff-c90dac2b22039f287a63d2212d415a2d1d9aae1d8d8ba69eb034ba1ad0c00b4bR594).

pierre-gautier:
Thanks for your answer.
As I said before I'm actually quite OK with this philosophy.
But I need to confirm that you are aware about the fact that it is an important behavior change and upgrading from 4.1.5 to 4.2.0 will cause NPEs in most applications using initialized collections, are you OK with that ?

mp911de:
We should have introduced this change much earlier as the change is more honest. You can always intercept values in a constructor or, if you use `@Access(PROPERTY)`, then you can intercept collection `null` values in setters.

Alternatively, you can make use of lifecycle events to post-process the `Document` values that are `null` for a collection.

PR #4574, opened 2023-11-30
Document list/map/set initialization on read.

christophstrobl:
Update the reference documentation about collection initialization on read.
Add required tests to make sure it behaves as expected and simplify `BeanUtils` value presence check.

See: #4571


mp911de:

Issue #2836, opened 2018-04-20
Indexed affects not only root collection [DATAMONGO-1965]


spring-projects-issues:
**[Vadzim Parafianiuk](https://jira.spring.io/secure/ViewProfile.jspa?name=proteinby)** opened **[DATAMONGO-1965](https://jira.spring.io/browse/DATAMONGO-1965?redirect=false)** and commented

If I use `@Indexed` in one class, I want spring to create Index only in collection for this class, but it also creates index in another classes that use this `@Indexed` class as embedded field. As a result, if I use "unique = true" it also affects embedded fields in another classes. See my question on stackoverflow with more rich description and example.

 

I've found a solution: I create ApplicationListener, and after context loading I use MongoTemplate to create indexes manually. But it is not convenient

---

**Reference URL:** https://stackoverflow.com/questions/49930576/spring-data-mongodb-indexedunique-true-in-inner-field

jasonkingatconversocial:
I have just been tripped up with this - it certainly feels like a bug for my use-case.
Extra index(es) are created when a class contains references to **itself**:

Example:
```
@Document("People")
class Person {
    @Indexed private String name;
    private Map<String, Person> friends = new HashMap<>();
}
```
The following indexes are created. `"name" : "friends.name"` is not wanted/desired.
```
> db.People.getIndexes()
[
	{
		"v" : 2,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "test.People"
	},
	{
		"v" : 2,
		"key" : {
			"name" : 1
		},
		"name" : "name",
		"ns" : "test.People"
	},
	{
		"v" : 2,
		"key" : {
			"friends.name.name" : 1
		},
		"name" : "friends.name",
		"ns" : "test.People"
	}
]
```

This can quickly get out of hand creating many unwanted indexes if the class contains several `@Indexed` fields.

@odrotbohm - please let me know if you would like me to report a new issue specifically for this case.

rolag-it:
Hello, 
I just submit a PR with a simple proposal: an annotation `@IgnoreIndexes` to put on fields mapped as embedded document. 
Adding this annotation, the indexes defined in classes of embedded fields will not be created on root collection.

mp911de:
[...]

PR #4600, opened 2023-12-28
Add annotation @IgnoreIndexes

rolag-it:
Allow to ignore indexes defined on class of properties when mapped as embedded document

See: #2836


rolag-it:
The currently implemented resolution of annotation-based indexes for properties mapped as embedded document is quite odd resulting counter-intuitive and bug prone.

The new annotation `@IgnoreIndexes` allows to exclude unwanted indexes definitions when mapping embedded document; it is a low footprint, no side-effects, totally backward compatible solution that enables great flexibility to annotation-based indexes.

Also it facilitates the embedded document pattern, a key approach for mapping relation in MongoDB.

christophstrobl:
Thank you @rolag-it for taking the time to look into this issue and create the PR! 
I second @mp911de recent comment on #2836 regarding use of indexing annotations and do not think adding more annotations is the way to go here.

PR #4666, opened 2024-03-19
Do not set `AggregationOptions.allowDiskUse` by default

### LINKED ISSUE #4664: Aggregation `allowDiskUse` should not be `false` by default
Starting from version 6, Mongo has server parameter `allowDiskUseByDefault` which is `true` by default. In Spring `allowDiskUse` for aggregation pipelines is false by default (because of using `boolean`) which means Spring always implicitly overrides Mongo default beheviour. Queries have no such issue. I think Spring should not set any parameters silently overriding Mongo default behaviour

```java

public class AggregationOptions implements ReadConcernAware, ReadPreferenceAware {
...
private final boolean allowDiskUse;
...
}
```

mp911de:
With MongoDB changing its default, we no longer set `allowDiskUse` to false if the value is not configured.

Closes #4664

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.3.0-SNAPSHOT</version>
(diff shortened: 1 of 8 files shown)


christophstrobl:
LGTM

PR #4668, opened 2024-03-19
Added include() and exclude() methods that accept Collection in field projection

### LINKED ISSUE #4625: Allow query projection methods `.include()` and `.exclude()` to accept `Collection`
Hi, we have a bunch of APIs that allow our users to specify a comma-separated list of fields that they want to receive back from the API response, which lowers our network bandwidth usage:
```java
@GetMapping("/books")
public List<Book> getAllBooks(
	@RequestParam(defaultValue = "") Set<String> fieldsToInclude     //Set removes duplicate fields
) {
	return bookDAO.getBooks(fieldsToInclude)
}
```

Currently, spring-data allows us to project fields on a query using the `include()` and `exclude()` methods. However, the `include()` method only accepts `String` or `String...`. Thus, we're forced to add code to convert our Lists into String arrays just to satisfy the `include()` function's signature.
```java
public List<Book> getBooks(Collection<String> fieldsToInclude) {
    var query = new Query();
    query.fields().include(fieldsToInclude.toArray(new String[0]));
    query.with(Sort.by("date").descending());
    return mongoTemplate.find(query, Book.class);
}
```
This adds complexity and maintenance overhead even though `Collection` is the more flexible and more common data structure to use in Java code over `String[]`.

**So, I would like to propose adding a new version of `include` and `exclude` that accepts `Collection`** so that we can directly use modern `Collection` implementors (List, Set, etc) for these methods instead of being forced to convert them to primitive `String[]` first?

---

The implementation would be pretty simple, so this should be an easy upgrade that would greatly improve the public API of  spring-data-mongo `Query` objects to make them more flexible and usable:
```java
public Field include(Collection<String> fields) {
	Assert.notNull(fields, "Keys must not be null!");
	for (String key : fields) {
		criteria.put(key, 1);
	}
	return this;
}
// similar for exclude
```

(I can raise a PR for this if that's easier for the spring team)

w0r1t:
Hello! This PR fix #4625. This is my first PR, please let me know if I've left anything out 🙏

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Field.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Field.java
@@ -16,6 +16,7 @@
 package org.springframework.data.mongodb.core.query;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -35,6 +36,7 @@ import org.springframework.util.ObjectUtils;
  * @author Christoph Strobl
  * @author Mark Paluch
  * @author Owen Q
+ * @author Kirill Egorov
  */
 public class Field {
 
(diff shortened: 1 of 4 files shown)


christophstrobl:
Thank you @w0r1t for taking the time to create this PR! Well have a look :)

christophstrobl:
Thank you @w0r1t for your contribution - merged to _main_ development line via d2a9744.

PR #4679, opened 2024-04-04
Add logging a sort options for find queries

KSH-code:
If use a querydsl with mongo such as
```
from(collection)
            .where(predicate)
            .orderBy(orderField.asc())
            .offset(pageable.offset)
            .limit(pageable.pageSize.toLong())
            .fetch()
```
and also log is enabled,

Then this query logs like
```
[C] 2024-04-04T16:06:13.855+09:00|DEBUG|pool-1-thread-1 @coroutine#3||MongoTemplate.java:2691|find using query: { "updatedAt" : { "$gte" : { "$date" : "2024-04-03T07:06:13.829Z"}, "$lte" : { "$date" : "2024-04-04T07:06:13.829Z"}}} fields: Document{{}} for class: class blahblah in collection: Collection
```

Thus it needs to be logged with sort query.

close https://github.com/spring-projects/spring-data-mongodb/issues/4686


KSH-code:
@pivotal-cla This is an Obvious Fix

mp911de:

Issue #2764, opened 2018-02-05
Support dynamic naming strategy for collection names [DATAMONGO-1863]


spring-projects-issues:
**[Dan Markhasin](https://jira.spring.io/secure/ViewProfile.jspa?name=minimizer)** opened **[DATAMONGO-1863](https://jira.spring.io/browse/DATAMONGO-1863?redirect=false)** and commented

As described in [DATAMONGO-525](https://jira.spring.io/browse/DATAMONGO-525) (https://jira.spring.io/browse/DATAMONGO-525), we are looking for a way to generate collection names dynamically to support multi-tenancy.
This can be achieved by something as simple as a prefix or by a naming strategy mechanism, simliar to FieldNamingStrategy.

For example, the most simple solution that would work (for us) is to have something like this:

```java
@Document(prefix = "${info.service_name}")
class Dog
{
private String breed;
}
```

That would generate a collection called animalService_Dog (assuming the service name is animalService).
Any other solution is welcome, but I believe this is a major feature that is currently lacking which is critical for multi-tenancy support

---

**Reference URL:** https://jira.spring.io/browse/DATAMONGO-525

1 votes, 3 watchers

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

alxalx22:
> Does that already do what you want?

No, we need a way to specify in `@Document` annotation the collection name as property from properties file

harinibodipudi:
https://www.examplefiles.net/cs/1040155

This link specifies how we can use dynamic collection names with the environment variables. It worked for me

Steve973:
I am trying to do something like this, too.  My use case entails a configuration that specifies a list of mongo urls, and/or a list of collections for the same URL.  If, in my application, I want to perform a kind of query federation over some configured number of collections in one or more mongo instances, I need a way to set up the collection name in the repositories that I configure, especially if they use the same model class, and I don't want to mandate the exact name of the collection.  Please let me know if that explanation doesn't make sense, and I will do my best to clear it up.

[...]

PR #4683, opened 2024-04-05
Add support for Value Expressions for Repository Query methods

mp911de:


marcingrzejszczak:
This will also fix https://github.com/spring-projects/spring-data-mongodb/issues/2764

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/AbstractReactiveMongoQuery.java
```diff
@@ -415,17 +463,65 @@ protected Mono<ParameterBindingDocumentCodec> getParameterBindingCodec() {
...
+							@Override
+							public <T> T evaluate(String expressionString) {
+
+								ValueExpression expression = valueExpressionDelegate.parse(expressionString);
+
+								return (T) expression.evaluate(evaluationContext);
+							}
+						};
+					});
+	}
```
christophstrobl, line 523: can we move the inner `ValueExpressionEvaluator` to a dedicated class that holds the evaluation context, so we can use for both?

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/util/json/ParameterBindingJsonReaderUnitTests.java
```diff
@@ -406,7 +410,7 @@ public void capturingExpressionDependenciesShouldNotThrowParseErrorForSpelOnlyJs
 		String json = "?#{ true ? { 'name': #name } : { 'name' : #name + 'trouble' } }";
 
 		new ParameterBindingDocumentCodec().captureExpressionDependencies(json, (index) -> args[index],
-				new SpelExpressionParser());
+				ValueExpressionParser.create(SpelExpressionParser::new));
```
christophstrobl, line 413: Might make sense to also add some tests to the binding parser and check if the current test setup is making use of value expressions already.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/repository/query/StringBasedMongoQueryUnitTests.java
```diff
@@ -293,10 +303,36 @@ public void shouldSupportExpressionsInCustomQueries() {
...
+		org.springframework.data.mongodb.core.query.Query query = mongoQuery.createQuery(accessor);
+		org.springframework.data.mongodb.core.query.Query reference = new BasicQuery("{'lastname' : 'bar'}");
+
+		assertThat(query.getQueryObject()).isEqualTo(reference.getQueryObject());
+	}
+
 	@Test // DATAMONGO-1244
 	public void shouldSupportExpressionsInCustomQueriesWithNestedObject() {
 
-		ConvertingParameterAccessor accessor = StubParameterAccessor.getAccessor(converter, true, "param1", "param2");
```
christophstrobl, line 299: why's `param2` gone?

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/repository/query/StringBasedMongoQueryUnitTests.java
```diff
@@ -293,10 +303,36 @@ public void shouldSupportExpressionsInCustomQueries() {
...
+	public void shouldSupportPropertiesInCustomQueries() {
+
+		properties.put("foo", "bar");
+		ConvertingParameterAccessor accessor = StubParameterAccessor.getAccessor(converter);
+		StringBasedMongoQuery mongoQuery = createQueryForMethod("findByQueryWithProperty");
+
+		org.springframework.data.mongodb.core.query.Query query = mongoQuery.createQuery(accessor);
+		org.springframework.data.mongodb.core.query.Query reference = new BasicQuery("{'lastname' : 'bar'}");
+
+		assertThat(query.getQueryObject()).isEqualTo(reference.getQueryObject());
```
christophstrobl, line 329: what do we expect when the property is missing and does not have a default?

christophstrobl (commented):
Sme `since` and `deprecation` tags are off and we should update the documentation.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/repository/query/StringBasedMongoQueryUnitTests.java
mp911de: because the method declares 2 parameters and previously we passed in 3 arguments.

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/repository/query/StringBasedMongoQueryUnitTests.java
marcingrzejszczak: IMO it should break (and it does right now, I'll add a test for that)

Issue #4687, opened 2024-04-11
Aggregation criteria match mapping fails with `NullPointerException`


piotrmucha:
spring-data-mongodb version: 4.2.4

I am trying to move following query from JSON to MongoTemplate:
```yaml
{
  '$match': {
      $expr: {
        $regexMatch: {
          input: {
            $toString:
              "fieldToConvert",
          },
          regex: "aa",
          options: "i",
        }
      }
  }
}
```
I have tried following approach:

```java
var toString = ConvertOperators.valueOf("fieldToConvert").convertToString();
var regexMatch = StringOperators.valueOf(toString).regexMatch("aa", "i");
var expr = Criteria.expr(regexMatch);
var match = Aggregation.match(expr);
var aggregation = Aggregation.newAggregation(match);
mongoTemplate.aggregate(aggregation, "collectionName", classInstance);
```

But it doesn't work, exception like this is thrown:
```
Cannot invoke "org.springframework.data.mongodb.core.mapping.MongoPersistentEntity.getType()" because "entity" is null
	at org.springframework.data.mongodb.core.convert.QueryMapper.convertSimpleOrDocument(QueryMapper.java:592) ~[spring-data-mongodb-4.2.4.jar:4.2.4]
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedKeyword(QueryMapper.java:411) ~[spring-data-mongodb-4.2.4.jar:4.2.4]
	at org.springframework.data.mongodb.core.convert.QueryMapper.getMappedObject(QueryMapper.java:130) ~[spring-data-mongodb-4.2.4.jar:4.2.4]
	at org.springframework.data.mongodb.core.aggregation.TypeBasedAggregationOperationContext.getMappedObject(TypeBasedAggregationOperationContext.java:82) ~[spring-data-mongodb-4.2.4.jar:4.2.4]
	at org.springframework.data.mongodb.core.aggregation.TypeBasedAggregationOperationContext.getMappedObject(TypeBasedAggregationOperationContext.java:77) ~[spring-data-mongodb-4.2.4.jar:4.2.4]
	at org.springframework.data.mongodb.core.aggregation.MatchOperation.toDocument(MatchOperation.java:74) ~[spring-data-mongodb-4.2.4.jar:4.2.4]
[...]

PR #4695, opened 2024-04-30
Fix NPE when rendering untyped aggregation operation as part of query.

christophstrobl:
Resolves: #4687


mp911de:

PR #4699, opened 2024-05-16
Use collection name instead type to look up the collection for index creation

### LINKED ISSUE #4698: MongoTemplate.indexOps(collectionName, type) ignores collectionName when type is not null
I have an object type that is stored in MongoDB under two tables: one using the default (class) table name, another using a custom name.  The code needs to ensure a geospatial index exists on the table before doing a “near” query.  It looks something like this:
```
        getMongoTemplate().indexOps("topPlaces", Place.class).ensureIndex(
                new GeospatialIndex("location")
                        .typed(GeoSpatialIndexType.GEO_2DSPHERE));

        GeoResults<Place> results = getMongoTemplate()
                .geoNear(nearQuery, Place.class, "topPlaces");
```
When executing the query, it fails with:

 > $geoNear requires a 2d or 2dsphere index, but none were found' on server localhost:27017

and checking the tables in mongosh I found the index was created on the “place” table, not on “topPlaces”.  I traced the fault down to this line in `DefaultIndexOperations.execute(callback)`:

https://github.com/spring-projects/spring-data-mongodb/blob/7821c2a545da5ac10529f0158639e998dfb0072a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/DefaultIndexOperations.java#L212

which prioritizes executing the callback for the `type` of the collection if present over the `collectionName`.

If this is the intended behavior of `indexOps(collectionName, type)`, it’s not documented.  The javadoc implies that the `type` parameter is only used for field mapping.  It also says that `collectionName` is required, even though it isn’t used.

_(Caveat: I’m currently running with spring-data-mongodb 4.0.12, but it appears this `execute` method has not been changed in the latest code here on GitHub.)_

mp911de:
Closes #4698

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.3.0-SNAPSHOT</version>
+	<version>4.3.0-GH-4698-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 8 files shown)


christophstrobl:
LGTM

Issue #4674, opened 2024-03-26
Criteria.regex is converted to String when applying to ID fields


spekdrum:
Hi, this issue #4649 is still happening in 4.2.4 **with ID fields**. 

In `QueryMapper#getMappedValue` method, `applyFiledTargetTypeHitToValue()` returns the Pattern object correctly after the fix, but later ends calling `QueryMapper#convertId()` method, which ends calling `getConversionService().canConvert()` in `MongoConverter.convertId()` method. This is returning true, so Pattern is converted to String anyway.

![image](https://github.com/spring-projects/spring-data-mongodb/assets/3244109/c4b8ff5d-09ca-4efb-9a51-0f3005e6f396)

![image](https://github.com/spring-projects/spring-data-mongodb/assets/3244109/89c6cf8a-3aad-493a-b42b-2b7b11c474da)

christophstrobl:
Thank you for getting in touch. It would help us investigating the issue if you could take the time to provide a complete [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) (something that we can unzip or git clone, build, and deploy) that reproduces the problem.

spekdrum:
Hi, I don't have the time right now for a full example, maybe next week.

Just create a query with a regexp on the key column, i.e.:

`		final Query query = new Query();

		query.addCriteria(where(ID_FIELD).regex("^3517-"));			

		return this.mongoTemplate.find(query, RateRoomPricingData.class, RATES_COLLECTION);
`

This wont return any value because the regexp expression is lost due to conversion.

spekdrum:
I built an example reproducing the problem. 
[mongo-demo.zip](https://github.com/spring-projects/spring-data-mongodb/files/15293870/mongo-demo.zip)

In summary, the issue is that when filtering the _id field of a collection with a regexp, it doesn't return results, because the resulting Pattern is being converted to a String anyway. I showed the key lines in a debugging season in my first post.

Using spring boot parent version 3.2.0+ the bug is reproduced, but using 3.1.11 it works.

I tried to enforce the spring-data-mongodb dependency version only, to 4.1.x, which is the working one, but the problem persists, so maybe it's out from this lib.


Code overview:

```
@SpringBootApplication
public class MongoDemoApplication {

	public static void main(final String[] args) {

		SpringApplication.run(MongoDemoApplication.class, args);
	}

	@Bean
[...]

PR #4718, opened 2024-06-10
Fix conversion of regular expression targeting id properties in queries 

christophstrobl:
This commit fixes an issue where patterns targeting id properties might have been falsely converted into the properties type, turning a `Pattern` into its `String` representation.

Closes: #4674


mp911de:

Issue #4707, opened 2024-05-27
Id conversion handling in `QueryMapper.getMappedValue(…)` doesn't consider nesting


antechrestos:
While upgrading to springboot 3.3.0, I detect a strange behaviour. 

I use `spring-boot-starter-data-mongodb-reactive` with `CoroutineCrudRepository`.

When using coposite id, I cannot save documents. Every save ends with error

```
Write operation error on server localhost:32811. Write error: WriteError{code=66, message='After applying the update, the (immutable) field '_id' was found to have been altered to _id: { user: { _id: "1", tenant: "tenant" }, demoId: "demo-id", timestamp: new Date(1716798131072) }', details={}}.
```

I made a [demo repository](https://github.com/antechrestos/spring-data-mongo-bug)

mp911de:
`convertIdField(…)` applies structures and types of the Id field even for nested documents if the Id is a document with nested documents. We have to fix that.

PR #4719, opened 2024-06-10
Fix id conversion for fields of complex id. 

christophstrobl:
This PR fixes an issue where the property type for nested fields of an complex id is not handed over correctly leading to wrong conversion results eg. for Instant types that got then turned into ObjectIds.

Closes: #4707


mp911de:

PR #4721, opened 2024-06-11
Apply type conversion to id types when serializing querydsl queries.

christophstrobl:
This PR makes sure to apply type conversion. 
However the disabled tests is somehow worrysome.


mp911de:

Issue #4722, opened 2024-06-11
Aggregation pipeline breaks if `$replaceRoot` stage is present


jmayday:
This is similar to already closed https://github.com/spring-projects/spring-data-mongodb/issues/4285. 

If I include `$replaceRoot` stage, aggregation fails with `CodecConfigurationException`. Below is the (bit anonymized) aggregation itself. Workaround is to use `$project` instead of `$replaceRoot` (which is not that convenient). 

```
org.bson.codecs.configuration.CodecConfigurationException: Can't find a codec for CodecCacheKey{clazz=class org.springframework.data.mongodb.core.aggregation.BooleanOperators$And, types=null}.
```

```
[
    {
        "$match": {
            "$and": [
                {
                    "fieldA": "some value"
                }
            ]
        }
    },
    {
        "$sort": {
            "updatedAt": -1
        }
    },
    {
        "$group": {
            "_id": "$_some_field_id",
            "latestRecord": {
                "$first": "$$ROOT"
            }
        }
    },
    {
        "$replaceRoot": {
            "newRoot": "$latestRecord"
        }
    },
    {
        "$lookup": {
            "from": "joinedCollectionName",
            "let": {
                "fieldB": "$fieldB",
                "fieldA": "$fieldA",
                "fieldC": "$fieldC"
            },
            "pipeline": [
                {
                    "$match": {
                        "$expr": {
                            "$java": org.springframework.data.mongodb.core.aggregation.BooleanOperators$And@5d77efc8
                        }
                    }
                }
            ],
            "as": "myAlias"
        }
    }
]
```

Some dependency versions:
```
[INFO] +- org.springframework.boot:spring-boot-starter-data-mongodb:jar:3.3.0:compile
[INFO] |  +- org.mongodb:mongodb-driver-sync:jar:5.0.1:compile
[INFO] |  |  +- org.mongodb:bson:jar:5.0.1:compile
[INFO] |  |  \- org.mongodb:mongodb-driver-core:jar:5.0.1:compile
[INFO] |  |     \- org.mongodb:bson-record-codec:jar:5.0.1:runtime
[INFO] |  \- org.springframework.data:spring-data-mongodb:jar:4.3.0:compile
[INFO] |     +- org.springframework:spring-tx:jar:6.1.8:compile
[INFO] |     \- org.springframework.data:spring-data-commons:jar:3.3.0:compile
```

christophstrobl:
[...]

PR #4723, opened 2024-06-13
Fix conversion of types when mapping Aggregation pipeline

christophstrobl:
This PR makes sure to apply conversion to non native mongo types when the context does not expose fields.

Resolves: #4722

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationRenderer.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationRenderer.java
@@ -62,7 +62,7 @@ class AggregationOperationRenderer {
 				if (operation instanceof InheritsFieldsAggregationOperation || exposedFieldsOperation.inheritsFields()) {
 					contextToUse = contextToUse.inheritAndExpose(fields);
 				} else {
-					contextToUse = fields.exposesNoFields() ? DEFAULT_CONTEXT
+					contextToUse = fields.exposesNoFields() ? ConverterAwareNoOpContext.instance(rootContext)
 							: contextToUse.expose(fields);
 				}
 			}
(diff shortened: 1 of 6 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/aggregation/AggregationOperationRenderer.java
```diff
@@ -62,7 +62,7 @@ static List<Document> toDocument(List<AggregationOperation> operations, Aggregat
 				if (operation instanceof InheritsFieldsAggregationOperation || exposedFieldsOperation.inheritsFields()) {
 					contextToUse = contextToUse.inheritAndExpose(fields);
 				} else {
-					contextToUse = fields.exposesNoFields() ? DEFAULT_CONTEXT
+					contextToUse = fields.exposesNoFields() ? ConverterAwareNoOpContext.instance(rootContext)
```
mp911de, line 65: I'm not sure I follow why we have this distinction here. Switching to `contextToUse = contextToUse.expose(fields);` also lets all the tests pass.

christophstrobl: it might cause the mapping to consider filenames of follow up stages to be object to field name mapping which should not be done. if no tests fail, we should make sure that we have coverage there and do not hit a blank spot.

mp911de: Okay, then this one needs to wait for the next service release as we supposedly do not have sufficient test coverage.

christophstrobl: confirmed, we're missing a test here - like if we have an operation that forces field mapping to back off (like the replaceRoot one) a strict context would fail because of the missing exposure.

moldo1997:
Hello @christophstrobl ,
Do you have an expected time for which this ticket will be merged?
Thanks,
Cristian

moldo1997:
Hello @christophstrobl ,

Are there any updates regarding this ticket?

Was the test necessary for the ticket to be merged added?

Thank you,
Cristian

PR #4728, opened 2024-06-18
Consider null value settings for types with custom conversion

christophstrobl:
This PR fixes an issue where settings regarding storage of `null` values had been ignored if a custom converter took care of the conversion.

Let's do an initial round of review and check how dbrefs should be treated when there's this kind of conversion. Also, if we decide to go for it, we need to update the documentation.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/DocumentAccessor.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/DocumentAccessor.java
@@ -92,6 +92,10 @@ class DocumentAccessor {
 
 		Assert.notNull(prop, "MongoPersistentProperty must not be null");
 
+		if (value == null && !prop.writeNullValues()) {
+			return;
+		}
+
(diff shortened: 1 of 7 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/DocumentAccessor.java
```diff
@@ -172,4 +172,21 @@ private static Document getOrCreateNestedDocument(String key, Bson source) {
...
 		return nested;
 	}
+
+	DocumentAccessor withCheckFieldMapping(boolean checkFieldMapping) {
+
+		if(!checkFieldMapping) {
+			return this;
+		}
+
+		return new DocumentAccessor(this.document) {
```
mp911de, line 182: If we want to handle field write checks here, how about creating a named subclass that we would use across `MappingMongoConverter` in the write paths?

CybAtax:
Hey guys, this PR looks like it would greatly fix the issue we'd have with the recent spring boot version. I was wondering, whether any update to this is planned? If support for this issue is desired, I'd love to help :)

PR #4734, opened 2024-07-01
Fixed the bug, a null able list does not work

sikandar:
If there is a nullable list with values, the implementation should check it instead of wrapping the nullable list (even if it has values) into another newly created list by the Criteria class.

### Existing
Code: `if (!filter.ids.isNullOrEmpty()) criteria.and("id").`in`(filter.ids) `
Output: `{"id" : { "$in" : [["667e8b8af76f17213e4d4280"]]}}` **Wrong**

Code: `if (!filter.ids.isNullOrEmpty()) criteria.and("id").`in`(filter.ids!!)` **Required !!**
Output:` {"id" :{ "$in" : ["667e8b8af76f17213e4d4280"]}}` **Correct**

### After Fix
Code: `if (!filter.ids.isNullOrEmpty()) criteria.and("id").`in`(filter.ids) ` **With and Without !!**
Output: `{"id" : { "$in" : ["667e8b8af76f17213e4d4280"]}}` **Correct**

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Criteria.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Criteria.java
@@ -317,7 +317,7 @@ public class Criteria implements CriteriaDefinition {
 	public Criteria in(Object... values) {
 		if (values.length > 0 && values[0] instanceof Collection) {
 			throw new InvalidMongoDbApiUsageException(
-					"You can only pass in one argument of type " + values[1].getClass().getName());
+					"You can only pass in one argument of type " + values[0].getClass().getName());
 		}
 		criteria.put("$in", Arrays.asList(values));
 		return this;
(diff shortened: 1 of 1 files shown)


christophstrobl:
Thank you for getting in touch. Please explain why you think this is necessary and provide throughout testcases for the proposed change since there's a method that accepts a `Collection`. 

As `{"$in" : [["n1", "n2"]]}` matches differently from `{"$in" : ["n1", "n2"]}` I think revisiting the arrangement makes sense.

sikandar:
> Thank you for getting in touch. Please explain why you think this is necessary and provide throughout testcases for the proposed change since there's a method that accepts a `Collection`.
> 
> As `{"$in" : [["n1", "n2"]]}` matches differently from `{"$in" : ["n1", "n2"]}` I think revisiting the arrangement makes sense.

This solution will be able to work with Variable Arguments

Issue #4744, opened 2024-07-16
@Query with sort does not work in Native Image (Unexpected AOP exception)


VR4J:
Hello 👋 

I'm trying to use the `@Query` annotation with sorting functionality like the one below.
```java
@Query(value = "{ 'forename': ?0 }", sort = "{ 'forename': 1 }")
Optional<Person> getFirstPersonByForename(String forename);
```
Although after creating a native image with `./gradlew bootBuildImage` the following runtime error is thrown.
```sh
Application run failed
org.springframework.aop.framework.AopConfigException: Unexpected AOP exception
        at org.springframework.aop.framework.CglibAopProxy.buildProxy(CglibAopProxy.java:236)
        at org.springframework.aop.framework.CglibAopProxy.getProxy(CglibAopProxy.java:163)
        at org.springframework.aop.framework.ProxyFactory.getProxy(ProxyFactory.java:110)
        at org.springframework.data.mongodb.repository.query.QueryUtils.decorateSort(QueryUtils.java:69)
        at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.applyAnnotatedDefaultSortIfPresent(AbstractMongoQuery.java:230)
        at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.doExecute(AbstractMongoQuery.java:138)
        at org.springframework.data.mongodb.repository.query.AbstractMongoQuery.execute(AbstractMongoQuery.java:119)
        at org.springframework.data.repository.core.support.RepositoryMethodInvoker.doInvoke(RepositoryMethodInvoker.java:170)
        at org.springframework.data.repository.core.support.RepositoryMethodInvoker.invoke(RepositoryMethodInvoker.java:158)
        at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.doInvoke(QueryExecutorMethodInterceptor.java:164)
        at org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.invoke(QueryExecutorMethodInterceptor.java:143)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
        at org.springframework.data.projection.DefaultMethodInvokingMethodInterceptor.invoke(DefaultMethodInvokingMethodInterceptor.java:70)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
[...]

PR #4747, opened 2024-07-17
Generate proxy and register reflection for default sorting during AOT.

christophstrobl:
This PR makes sure to generate required cglib proxies during AOT phase so they are ready to use within a native image. Prior to this change default sorting raised an error as class generation is not allowed in a GraalVM native image.

Resolves: #4744

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.4.0-SNAPSHOT</version>
+	<version>4.4.x-GH-4744-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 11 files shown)


mp911de:
Why do we even have the need for proxying the Query class? In all other places where we amend the query, we do this via proper methods. Did we consider (or why did we reject) something like `query.getSortObjectOr(() -> Document.parse(method.getAnnotatedSort()))`?

christophstrobl:
Some years back (when GraalVM had not released) using proxies allowed to sneak in the default sorting without changing the original sort object. Without changing internals there's no other way to support the arrangement for previous versions. Going forward it makes sense to work on a solution that can do without it.

mp911de:
That's merged and backported now. For `4.4.x`, we replaced query proxying with a proper query copy.

slowbreathing:
Getting the same error for spring-boot-starter-data-mongodb-reactive

christophstrobl:
@slowbreathing would you mind opening a new ticket providing a [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) (something that we can unzip or git clone, build, and deploy) that reproduces the problem.

slowbreathing:
@christophstrobl Done that. https://github.com/spring-projects/spring-data-mongodb/issues/4900

Issue #4745, opened 2024-07-17
Issue referencing or using field value set with SetWindowFieldsOperation shift


alex-ionescu-teamextension:
Hello,

I am using spring-data-mongodb-4.2.5.jar and I have the usecase where I am doing a shift to get the next or previous value of a timestamp field using $setWindowFields.
The problem with the Java approach is that the resulting field is not correctly referenced in the aggregation pipeline, resulting in a null value when doing a $dateDiff.

This is how I am doing the $setWindowFields / $dateDiff:

```
SetWindowFieldsOperation timestampWindowFields = SetWindowFieldsOperation.builder()
                .partitionByField("metaData.deviceId")
                .sortBy(Sort.by(Sort.Direction.ASC, "timestamp"))
                .output(DocumentOperators.valueOf("timestamp").shift(-1).defaultTo(-1))
                .as("previous")
                .build();

```
```
DateOperators.DateDiff dateDiff = DateOperators.zonedDateOf("timestamp", DateOperators.Timezone.valueOf("America/Chicago"))
                .diffValueOf("previous", DateOperators.TemporalUnit.from(ChronoUnit.SECONDS));
        SetOperation dateDiffOperation = set("timeDifference").toValue(dateDiff);

```
And this is how the operator ends up in the aggregation pipeline output:
![image](https://github.com/user-attachments/assets/a73de82a-b629-409a-8bee-e91a41133acd)

Any ideea why the "endDate" field is referenced like this, is there a way to overcome this issue to get the correct result? How could I do a dateDiff between computed / shifted fields? 

Thanks in advance,
Alex

christophstrobl:
thank you @alex-ionescu-teamextension for getting in touch. Please do not use images but take the time to provide a complete [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) (something that we can unzip or git clone, build, and deploy) that reproduces the problem.

alex-ionescu-teamextension:
Of course @christophstrobl ,

Given the below documents in a collection named "telemetry-debug":

```
[{
  "timestamp": {
    "$date": "2024-05-29T03:25:15.511Z"
  },
  "metaData": {
    "deviceId": "7FCTGAAA9PN023984"
  },
  "_id": {
    "$oid": "669639886b2b5d28d51866c3"
  }
},
{
  "timestamp": {
    "$date": "2024-05-29T03:25:15.651Z"
  },
  "metaData": {
[...]

PR #4751, opened 2024-07-25
Expose output of SetWindowFieldsOperation correctly to next aggregation stage.

christophstrobl:
Resolves: #4745


mp911de:

PR #4777, opened 2024-09-04
Support instantiation of Kotlin class with overridden read-only property

### LINKED ISSUE #4485: Problem with kotlin field with fixed value
I have the following data class in kotlin, where the "type" attribute always has the fixed value "BILLET", this is because the interface that my class implements asks for this attribute.

```kotlin
data class BilletTransaction(
    override val id: String,
    override val name: TransactionName,
    override val createdAt: ZonedDateTime,
    override val value: BigDecimal,
    val dueDate: LocalDate,
) : Transaction {
    override val type: TransactionType = TransactionType.BILLET
}
```

I have a record in the mongo database with this information, but when trying to read it I get the following error:

```
java.lang.UnsupportedOperationException: No accessor to set property private final transient <PACKAGE>.TransactionType <PACKAGE>.BilletTransaction.type
	at <PACKAGE>.BilletTransaction_Accessor_wa9qij.setProperty(Unknown Source)
	at org.springframework.data.mapping.model.InstantiationAwarePropertyAccessor.setProperty(InstantiationAwarePropertyAccessor.java:81)
	at org.springframework.data.mapping.model.ConvertingPropertyAccessor.setProperty(ConvertingPropertyAccessor.java:60)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.readProperties(MappingMongoConverter.java:606)
```

I noticed that in the class "InstantiationAwarePropertyAccessor" there is a method that tries to use a possible "setter" to configure the value, but it is not possible and the error occurs

image

Is it possible to get around this error?

mp911de:
Closes #4485

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.4.0-SNAPSHOT</version>
+	<version>4.4.0-4485-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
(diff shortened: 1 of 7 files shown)


christophstrobl:
LGTM

PR #4778, opened 2024-09-05
Fix missing query options when calling MongoTemplate#findAndModify

### LINKED ISSUE #4776: maxTimeMs in MongoTemplate.findAndModify(...) is not being honored.
JDK: Amazon Coretto 17.0.12
MongoDB version: 7.0.14
spring-data-mongodb version: 4.3.3

Sample code:
```
Criteria criteria = Criteria.where("name").is(name);
Query query = Query.query(criteria);
query.maxTimeMsec(5L);
Update update = Update.update("lastFetchedAt", new Date());
FindAndModifyOptions findAndModifyOptions = FindAndModifyOptions.options().returnNew(true);
return mongoTemplate.findAndModify(query, update, findAndModifyOptions, ExampleModel.class);
```

From the above, it is expected that the query to fetch the document should time out after 5ms, but this is not what happens. The maxTimeMsec set is ignored, and the document gets updated. 

If the following command is run through mongosh:

`db.runCommand({findAndModify: "example", query: {"name": "example100000"}, update: {"lastFetchedAt": new Date()}, maxTimeMS: 5})`

we get the following error:

`MongoServerError[MaxTimeMSExpired]: Plan executor error during findAndModify :: caused by :: operation exceeded time limit`

This is similar to issue #4374, but here maxTimeMs is not being carried forward to _UpdateContext_.

sriram-1729:
Forward maxTimeMsec from `Query` to `FindOneAndUpdateOptions`.

Closes #4776

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
@@ -181,6 +181,7 @@ import com.mongodb.client.result.UpdateResult;
  * @author Bartłomiej Mazur
  * @author Michael Krog
  * @author Jakub Zurawa
+ * @author Kinjarapu Sriram
  */
 public class MongoTemplate
 		implements MongoOperations, ApplicationContextAware, IndexOperationsProvider, ReadPreferenceAware {
@@ -1088,7 +1089,7 @@ public class MongoTemplate
 			operations.forType(entityClass).getCollation(query).ifPresent(optionsToUse::collation);
 		}
 
-		return doFindAndModify(createDelegate(query), collectionName, query.getQueryObject(), query.getFieldsObject(),
+		return doFindAndModify(createDelegate(query), collectionName, query, query.getFieldsObject(),
 				getMappedSortObject(query, entityClass), entityClass, update, optionsToUse);
 	}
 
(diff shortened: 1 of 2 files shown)


christophstrobl:
Thank you @sriram-1729 for taking the time to create this PR. 
The signature change of `doFindAndModify` may cause trouble for people upgrading. I'm not sure if there's away around this change. Need to have a closer look.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
```diff
@@ -1088,7 +1089,7 @@ public <T> T findAndModify(Query query, UpdateDefinition update, FindAndModifyOp
 			operations.forType(entityClass).getCollation(query).ifPresent(optionsToUse::collation);
 		}
 
-		return doFindAndModify(createDelegate(query), collectionName, query.getQueryObject(), query.getFieldsObject(),
+		return doFindAndModify(createDelegate(query), collectionName, query, query.getFieldsObject(),
```
sriram-1729, line 1092: In case this change is allowed, should I remove `query.getFieldsObject()`, since `query` is already being passed?

sriram-1729: I've changed the signature to remove redundant arguments to `doFindAndModify`. If the signature changes are to be kept to a minimum, I'll remove this commit.

sriram-1729:
Hello @christophstrobl! In case the method signature change is allowed, should I keep the changes minimal, or can I make further modifications to remove redundant arguments?

PR #4784, opened 2024-09-12
Fails fast if mongo not running

### LINKED ISSUE #4783: Fail the build when Mongo not running and tests are active
Mongo is required to be running for Spring Data Mongo DB tests to pass. It would be beneficial to fail the build when `skip tests` is NOT turned on. In other words if someone is skipping the tests then we don't care if Mongo is running or not. 

Additional note: Mongo must be running without a custom username and password, otherwise the tests will fail due to connection being unauthorized

marcingrzejszczak:
By using the maven download plugin we are trying to ping Mongo under 27017 port. This will be done once per build after each `clean` of the project (if the file has already been downloaded). This will not however check if the Mongo database is not having custom username and password.

fixes #4783

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.2.10-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


schauder:
I removed the milestone, since this doesn't seem to be done, I we want to do the release now.

PR #4790, opened 2024-09-16
Compound Wildcard Indexes

marcingrzejszczak:


marcingrzejszczak:
Question about the current implementation

A ccording to the [tests](https://github.com/spring-projects/spring-data-mongodb/pull/4790/files#diff-9d6eb6baf2625219a7e9a3a5248e2b8493187645dd8508bc4ebac3d29ab6d001) and [the implementation](https://github.com/spring-projects/spring-data-mongodb/pull/4790/files#diff-5dade87a1cd486bb401edd8c5b8f0c1fac37d7dd82c5fe84a805c00ac7e838d6R160-R173) the following combinations are NOT allowed (a `MappingException` is being thrown)
* `wildcardFieldName` NOT empty and `wildcardProjection` NOT empty
* `wildcardFieldName` EMPTY and `wildecardProjection` EMPTY

While the first condition is [correct](https://www.mongodb.com/docs/manual/core/indexes/index-types/index-wildcard/reference/restrictions/#compound-wildcard-index-restrictions) I have doubts about the second one. 

If `wildcardFieldName` is empty, that implies the default `$**` value. That in effect means that all fields are scanned and `wildcardProjection` is optional. Why is an exception being thrown? 

cc @christophstrobl

christophstrobl:
Yeah, you can create the index using an all fields term without narrowing down the scope via projections.
Would it make sense to switch the default value for `wildcardFieldName` to `$**` and replace the `empty` checks with something like an `isAllFieldsIndexTerm(...)` so we can get the code path there more expressive?

Another thing that crossed my mind is that you can specify more than one wildcard index on a collection - I guess same is true for compound wildcard ones, which would require the annotation to be repeatable.

marcingrzejszczak:
I looked into the code and re:

> Would it make sense to switch the default value for wildcardFieldName to $** and replace the empty checks with something like an isAllFieldsIndexTerm(...) so we can get the code path there more expressive?

If we look at `@CompoundIndex` it has the `def` method

> String def() default "";

and the Javadoc states 

> 	 * If left empty on nested document, the whole document will be indexed.

Which means that if we go with the approach that the `wildcardFieldName` would be equal to `$**` then I think we would be inconsistent. WDYT?

christophstrobl:
I do not agree - we've got the `fields` attribute that translates to the `def` of the compound index. So if that is blank and the default for `wildcardFieldName` is `$**` the entire doc would be indexed, right?

marcingrzejszczak:
OK I understand. For future reference we're talking about `@CompoundWildcardIndexed` that has ` fields` method that delegates to `@CompoundIndex`'s `def` method. Now if `@CWI` is empty which is the default that would be that `@CI` def is empty which means that it translates to `$**` which indexes the whole doc.

christophstrobl:
Does it make sense to provide a `CompoundWildcardIndexes` container to allow repeatable usage? Would also be good to have a test that is using value expressions for the index name to see if resolving those work as expected.

Issue #4686, opened 2024-04-09
Add a debug information for sort fields when using findByQuery


KSH-code:
When logging on debug environment, sort field is unsupported with `findBy`.
ref: https://github.com/spring-projects/spring-data-mongodb/pull/4679

PR #4794, opened 2024-09-25
Feature/sort logging

infoShare:
Incorrect sort logging introduced on https://github.com/spring-projects/spring-data-mongodb/issues/4686.
Logging query instead of sort.

@mp911de Could you please verify ?


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
```diff
@@ -2559,8 +2561,8 @@ protected <T> T doFindOne(String collectionName, CollectionPreparer<MongoCollect
 	 * @return the List of converted objects.
 	 */
 	protected <T> List<T> doFind(String collectionName, CollectionPreparer<MongoCollection<Document>> collectionPreparer,
-			Document query, Document fields, Class<T> entityClass) {
-		return doFind(collectionName, collectionPreparer, query, fields, entityClass, null,
+			Document query, Document fields, Document sort, Class<T> entityClass) {
```
mp911de, line 2564: We cannot simply change the signature of our API. With sort being an essential part for query logging, we might consider reworking `QueryCursorPreparer` to remove sorting from the preparer and putting it into method signatures.

infoShare: @mp911de - this PR was just to show what is the problem. The easiest way was to show what was incorrectly logged.

PR #4796, opened 2024-09-26
Update MongoDB Server compatibility matrix.

christophstrobl:
This PR adds MongoDB 8.0 to ci pipeline and updates the compatibility matrix to reflect tested server versions.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -134,6 +134,25 @@ pipeline {
 						}
 					}
 				}
+				stage('Publish JDK (Java 17) + MongoDB 8.0') {
+					when {
+						anyOf {
+							changeset "ci/openjdk17-mongodb-8.0/**"
+							changeset "ci/pipeline.properties"
+						}
+					}
+					agent { label 'data' }
+					options { timeout(time: 30, unit: 'MINUTES') }
+
+					steps {
+						script {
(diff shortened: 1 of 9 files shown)


ci/pipeline.properties
```diff
@@ -11,6 +11,7 @@ docker.mongodb.4.4.version=4.4.25
 docker.mongodb.5.0.version=5.0.21
 docker.mongodb.6.0.version=6.0.10
 docker.mongodb.7.0.version=7.0.2
+docker.mongodb.8.0.version=8.0.0
```
mp911de, line 14: We need to introduce that property in https://github.com/spring-projects/spring-data-build/blob/main/ci/pipeline.properties as well.

Jenkinsfile
```diff
@@ -134,6 +134,25 @@ pipeline {
 						}
 					}
 				}
+				stage('Publish JDK (Java 17) + MongoDB 8.0') {
```
mp911de, line 137: Looks like the new indents are spaces whereas the remaining file uses tabs.

ci/pipeline.properties
christophstrobl: yes, once this is approved, I'll add it to data-build.

PR #4807, opened 2024-10-11
Retain Regex options from the parsed `JsonToken`

### LINKED ISSUE #4806: `ParameterBindingJsonReader` doesn't consider regex options
It looks like the following query doesn't take the options "i" into account
@Query = { 'a': 'b', 'c': /^true$/i }

in the logs = { "a" : "b", "c" : { "$regularExpression" : { "pattern" : "^true$", "options" : ""}}}

mp911de:
We now retain `BsonRegularExpression.options` when resolving bind values from the original `BsonRegularExpression`.

Closes #4806

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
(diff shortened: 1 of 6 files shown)


christophstrobl:
LGTM 👍

Issue #4804, opened 2024-10-09
`NoSuchTransaction` in reactive MongoDB client when working with transactions


fremarti:
## Setup

Spring Boot `3.3.4` with `org.springframework.boot:spring-boot-starter-webflux` and `org.springframework.boot:spring-boot-starter-data-mongodb-reactive`. MongoDB version is `6.0.18`.

## MongoDB Config

Reactive mongo client is configured in configuration to activate transactional feature in mongo templates:

```kotlin
@Configuration
@EnableConfigurationProperties(MongoProperties::class)
class MongoConfiguration(
    private val mongoProperties: MongoProperties,
) : AbstractReactiveMongoConfiguration() {
    ...
    @Bean
    fun transactionManager(
        factory: ReactiveMongoDatabaseFactory?,
        properties: MongoProperties,
    ): ReactiveMongoTransactionManager {
        return ReactiveMongoTransactionManager(
            factory!!,
            TransactionOptions.builder().readPreference(ReadPreference.valueOf(properties.readPreference)).build(),
        )
    }
    ...
```

## Docker Setup

For local and integration testing a mongodb is configured using docker compose. The db is configured as single node replica set. The here mentioned init script just runs `rs.initiate(...)` to register replica set. In the application properties the according connection string is set with `mongodb://localhost:27017/?replicaSet=rs0`.

```yml
services:
  mongo:
    image: mongo:6.0.18
    ports:
      - "27017:27017"
    volumes:
      - ./bin/mongodb-init-replica-set.sh:/docker-entrypoint-initdb.d/mongodb-init-replica-set.sh:ro
    command: ["mongod", "--replSet", "rs0", "--bind_ip_all"]
  ...
```

## Application Code

I have an endpoint `PUT /foo` which should update multiple entries in a single collection. This update should be transactional. Before updating the entries, all entries are fetched by ids and some validation is done before updating the entries:

```kotlin
// FooController.kt
@RestController
class FooController(private val fooUseCase: FooUseCase) {
    ...
    @Transactional(label = ["mongo:readPreference=PRIMARY"])
    @PutMapping(
        value = ["/foo"],
        consumes = [MediaType.APPLICATION_JSON_VALUE],
    )
    fun foo(@RequestBody request: RequestDto): Mono<Void> {
        return fooUseCase
            .process(request)
            .doOnError { error ->
                logger.error("Failed", error)
            }
    }
}

// FooUseCase.kt
@Service
[...]

PR #4824, opened 2024-10-31
Retain order doing reactive save operations with multiple elements.

christophstrobl:
Resolves: #4804


fremarti:
Imho `flatMapSequentially` should be replaced with `concatMap` in order to preserve the order of execution (see comment https://github.com/spring-projects/spring-data-mongodb/issues/4804#issuecomment-2451666256)

PR #4831, opened 2024-11-10
Fix same hashCodes an different queries with include and exclude

### LINKED ISSUE #4032: Different Queries, same hashCodes
Consider two **org.springframework.data.mongodb.core.query.Query** instances almost identical except for one projection field which is 

- included in one query
- excluded in the other query

then the following test fails
```
Assertions.assertNotEquals(queryWithExclude.hashCode(), queryWithInclude.hashCode());
```

**Unit Test**
```
    @Test
    public void hashMapQueryTest() {
        Query queryWithExclude = new Query();
        queryWithExclude.fields().exclude("key1");
        queryWithExclude.fields().exclude("key2");

        Query queryWithInclude = new Query();
        queryWithInclude.fields().include("key1");
        queryWithInclude.fields().include("key2");

        Assertions.assertNotEquals(queryWithExclude.hashCode(), queryWithInclude.hashCode());
    }
```

**Comments**
The problem can be drilled down to the Field.hashCode()
```
ObjectUtils.nullSafeHashCode(this.criteria);
```

which seems to fails due to the HashMap implementation itself as proved by the following test
```
    @Test
    public void hashMapZeroOneTest() {
        HashMap<String, Integer> fieldZero = new HashMap<>();
        fieldZero.put("key1", 0);
        fieldZero.put("key2", 0);

        HashMap<String, Integer> fieldOne = new HashMap<>();
        fieldOne.put("key1", 1);
        fieldOne.put("key2", 1);

        Assertions.assertNotEquals(fieldZero.hashCode(), fieldOne.hashCode());
    }
```
however tests with other numbers than 0 or 1 works.

To conclude, the problem seem related to the use of 0 and 1 in the Field.criteria which then are translated according to the mongo projection syntax. Possible solution could be to replace in the Field class 0/1 used by include/exclude with enum INCLUDE/EXCLUDE values and consequently modify up to various get[Query|Field]Object().

Comments?

Regards,
Maurizio

13wjdgk:
Closes #4032

There was a bug where the hashCodes of the exclude and include queries were identical.

The issue was caused by the hashCode calculation in HashMap, where the XOR operation between the hashCode of the key and the hashCode of the value (0 or 1) resulted in duplicate hashCodes.

image

Since Exclude and Include are differentiated by 0 and 1 and are put into the hashMap, they were vulnerable to returning duplicate hashCodes.

To resolve this, the fix involved combining the key and value of the HashMap into a single String and calculating the hashCode based on that.

### DIFF
--- a/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/query/FieldUnitTests.java
+++ b/spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/query/FieldUnitTests.java
@@ -28,6 +28,7 @@ import java.util.List;
  * @author Owen Q
  * @author Mark Paluch
  * @author Kirill Egorov
+ * @author GaEun Kim
  */
 class FieldUnitTests {
 
@@ -84,5 +85,19 @@ class FieldUnitTests {
 		Field right = new Field().exclude(List.of("foo", "bar"));
 
 		assertThat(left).isEqualTo(right);
+	}
+
+	@Test
+	void assertDifferentHashCodesForExcludeAndIncludeQueries() {
+
+		Query queryWithExclude = new Query();
+		queryWithExclude.fields().exclude("key1");
+		queryWithExclude.fields().exclude("key2");
+
+		Query queryWithInclude = new Query();
+		queryWithInclude.fields().include("key1");
+		queryWithInclude.fields().include("key2");
+
+		assertThat(queryWithExclude.hashCode()).isNotEqualTo(queryWithInclude.hashCode());
 	}
 }
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Field.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/query/Field.java
@@ -37,6 +37,7 @@ import org.springframework.util.ObjectUtils;
  * @author Mark Paluch
  * @author Owen Q
  * @author Kirill Egorov
+ * @author GaEun Kim
  */
 public class Field {
 
@@ -286,7 +287,7 @@ public class Field {
 	@Override
 	public int hashCode() {
 
-		int result = ObjectUtils.nullSafeHashCode(criteria);
+		int result = ObjectUtils.nullSafeHashCode(criteria.toString());
 		result = 31 * result + ObjectUtils.nullSafeHashCode(slices);
 		result = 31 * result + ObjectUtils.nullSafeHashCode(elemMatches);
 		result = 31 * result + ObjectUtils.nullSafeHashCode(positionKey);


13wjdgk:
Instead of toString(), I also considered the following solution.

```java
	int result = nullSafeHashCodeForCriteria(criteria);
	
	public static int nullSafeHashCodeForCriteria(@Nullable Object obj) {
		if (obj == null) {
			return 0;
		} else if(obj instanceof Map<?,?>){
			int result = 0;
			for(Entry<?, ?> entry : ((Map<?, ?>)obj).entrySet()){
				if(entry.getValue() instanceof Integer){
					Integer value = (Integer)entry.getValue();
					int valueHashCode;
					if(value == 0){
						valueHashCode = FieldProjection.EXCLUDE.hashCode();
					}
					else if(value == 1){
						valueHashCode = FieldProjection.INCLUDE.hashCode();
					}else{
						valueHashCode = value.hashCode();
					}
					result += entry.getKey().hashCode() ^ valueHashCode;
				}else {
					result += entry.hashCode();
				}
			}
			return result;
		}

		return obj.hashCode();

	}
```

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/query/FieldUnitTests.java
```diff
@@ -85,4 +86,22 @@ void overriddenExclusionMethodsCreateEqualFields() {
...
+		queryWithExclude.fields().exclude("field1");
+		queryWithExclude.fields().exclude("field2");
+
+		Query queryWithInclude = new Query();
+		queryWithInclude.fields().include("key1");
+		queryWithInclude.fields().include("key2");
+		queryWithInclude.fields().include("field1");
+		queryWithInclude.fields().include("field2");
+
+		assertThat(queryWithExclude.hashCode()).isNotEqualTo(queryWithInclude.hashCode());
```
13wjdgk, line 105: Before the fix, the test above failed.
image

spring-data-mongodb/src/test/java/org/springframework/data/mongodb/core/query/FieldUnitTests.java
```diff
@@ -85,4 +86,18 @@ void overriddenExclusionMethodsCreateEqualFields() {
...
+
+		Query queryWithExclude = new Query();
+		queryWithExclude.fields().exclude("key1");
+		queryWithExclude.fields().exclude("key2");
+
+		Query queryWithInclude = new Query();
+		queryWithInclude.fields().include("key1");
+		queryWithInclude.fields().include("key2");
+
+		assertThat(queryWithExclude.hashCode()).isNotEqualTo(queryWithInclude.hashCode());
```
13wjdgk, line 101: Before the fix, the test above failed.
image

PR #4909, opened 2025-02-28
Update mapping-index-management.adoc

kdparkinson:
added clarity around @CompoundIndexes use

- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [X] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [X] You submit test cases (unit or integration tests) that back your changes.
- [X] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/src/main/antora/modules/ROOT/pages/mongodb/mapping/mapping-index-management.adoc
+++ b/src/main/antora/modules/ROOT/pages/mongodb/mapping/mapping-index-management.adoc
@@ -118,20 +118,24 @@ public class Person {
 
 [TIP]
 ====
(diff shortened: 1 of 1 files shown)


mp911de:
This should be common Java knowledge and so we do not want to document repeated annotation usage through Java's container mechanism. In any case, thanks for reaching out.

PR #4910, opened 2025-03-02
Introduce `ValidatingEntityCallback` and deprecate `ValidatingMongoEventListener`

rfelgent:

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingEntityCallback.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingEntityCallback.java
@@ -15,51 +15,52 @@
  */
 package org.springframework.data.mongodb.core.mapping.event;
 
-import java.util.Set;
-
 import jakarta.validation.ConstraintViolationException;
 import jakarta.validation.Validator;
-
+import java.util.Set;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
+import org.bson.Document;
+import org.springframework.core.Ordered;
 import org.springframework.util.Assert;
 
 /**
- * javax.validation dependant entities validator. When it is registered as Spring component its automatically invoked
- * before entities are saved in database.
+ * javax.validation dependant entities validator.
+ * <p>
+ * When it is registered as Spring component its automatically invoked
+ * after any {@link AbstractMongoEventListener} and before entities are saved in database.
  *
- * @author Maciej Walkowiak
- * @author Oliver Gierke
- * @author Christoph Strobl
+ * @author original authors of {@link ValidatingMongoEventListener}
+ * @author Rene Felgenträger
  *
- * @see {@link ValidatingEntityCallback}
+ * @see {@link ValidatingMongoEventListener}
  */
-public class ValidatingMongoEventListener extends AbstractMongoEventListener<Object> {
+public class ValidatingEntityCallback implements BeforeSaveCallback<Object>, Ordered {
 
-	private static final Log LOG = LogFactory.getLog(ValidatingMongoEventListener.class);
+	private static final Log LOG = LogFactory.getLog(ValidatingEntityCallback.class);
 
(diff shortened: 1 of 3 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingEntityCallback.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.bson.Document;
+import org.springframework.core.Ordered;
+import org.springframework.util.Assert;
+
+/**
+ * JSR-303 dependant entities validator.
+ * <p>
+ * When it is registered as Spring component its automatically invoked
+ * after any {@link AbstractMongoEventListener} and before entities are saved in database.
```
rfelgent, line 32: @mp911de , I copied the documentation from `ValidatingMongoEventListener` and adjusted it for the `ValidatingEntityCallback`. 
Is the statement regarding automatical invocation correct ?

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingEntityCallback.java
```diff
@@ -0,0 +1,78 @@
...
+ * @author original authors of {@link ValidatingMongoEventListener}
+ * @author Rene Felgenträger
+ *
+ * @see {@link ValidatingMongoEventListener}
+ */
+public class ValidatingEntityCallback implements BeforeSaveCallback<Object>, Ordered {
+
+	private static final Log LOG = LogFactory.getLog(ValidatingEntityCallback.class);
+
+	// TODO: create a validation handler (similar to "AuditingHandler") an reference it from "ValidatingMongoEventListener" and "ValidatingMongoEventListener"
```
rfelgent, line 43: personally, I could live with this implementation (and duplicate the validation code logic of `ValidatingMongoEventListener`)

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingEntityCallback.java
```diff
@@ -0,0 +1,78 @@
...
+	public ValidatingEntityCallback(Validator validator) {
+		Assert.notNull(validator, "Validator must not be null");
+		this.validator = validator;
+	}
+
+	// TODO: alternatively implement the "BeforeConvertCallback" interface and set the order to highest value ?
+	@Override
+	public Object onBeforeSave(Object entity, Document document, String collection) {
+
+		if (LOG.isDebugEnabled()) {
```
rfelgent, line 60: pls tell me, is "debug" or "trace" the appropriate choice ?

mp911de (commented):
Thanks a lot. For validation, we should also consider the reactive flow that should not throw exceptions but rather return `Mono.error(e)`. I think it makes sense to extract a tiny superclass returning `Set<ConstraintValitation>` and let the callback and event listener reuse those bits. We could then get rid of logging in each component.

Admittedly, it's not much code, but a few concerns that we do not require in subclasses. 

Also, we should deprecate the event listener and ideally find a place in the reference documentation to mention Bean Validation support.

Let me know how comfortable you are addressing these aspects and what guidance I can provide you with.

rfelgent:
>  ... I think it makes sense to extract a tiny superclass returning Set<ConstraintValitation>...

agreed. I will add a parent class and make `ValidatingEntityCallback` and `ValidatingMongEventListener` a subclass of it.

> ...  we should also consider the reactive flow ...

agreed, should the introduction of a reactive-based entity validor be part of this pull request or another one?

> ... we should deprecate the event listener ...

agreed, should this be part of this pull request or another one ?

mp911de:
Let's handle everything right here, no need to spread this enhancement over multiple tickets as these changes sum up nicely in a single logical change.

mp911de:
Thank you for your contribution. That's merged and polished now. I extracted a few bits and tweaked the documentation slightly, adding a reactive callback variant.

Issue #4918, opened 2025-03-10
Update using $inc does not work if @Version is present in the entity


symphony-enrico:
Given following entity:
```java
@Document("entitlementJobExecution")
@Data
@Builder
@AllArgsConstructor
@NoArgsConstructor
public class EntitlementJobExecution {
  @Id
  private ObjectId id;

  private Long executedLines;
  private Long successfullyExecutedLines;

  @Version
  private Long version;

}
```

and the following repository:
```java
@Repository
public interface EntitlementJobExecutionRepository extends MongoRepository<EntitlementJobExecution, ObjectId> {
  @Update("{ '$inc' : { 'executedLines' : ?1, 'successfullyExecutedLines' : ?2 } }")
  void findAndIncrementExecutedLinesAndSuccessfullyExecutedLinesById(
    ObjectId id,
    int executedLines,
    int successfullyExecutedLines
  );
}
```

The query does not work as expected. No error is reported, but it generates:
```
o.s.data.mongodb.core.MongoTemplate      : Calling update using query: { "_id" : { "$oid" : "67cf20da7b8c784443a05eb5"}} and update: { "$inc" : { "version" : 1}} in collection: entitlementJobExecution
```

Thus, executedLines and successfullyExecutedLines are not updated.
If I remove @Version from entity, it works as expected.

Thanks

symphony-enrico:
For info, generating manually the update with MongoTemplate and criteria:

```java
  public void incrementExecutedLinesAndSuccessfullyExecutedLinesById(
    ObjectId id,
    int executedLines,
    int successfullyExecutedLines
  ) {
    Query query = Query.query(Criteria.where("_id").is(id));
    Update update = new Update()
      .inc("executedLines", executedLines)
      .inc("successfullyExecutedLines", successfullyExecutedLines);
    mongoTemplate.updateFirst(query, update, EntitlementJobExecution.class);
  };
```
  
Works as expected:
```
o.s.data.mongodb.core.MongoTemplate      : Calling update using query: { "_id" : { "$oid" : "67d00a70a522b64eb464ddb6"}} and update: { "$inc" : { "executedLines" : 1, "successfullyExecutedLines" : 0, "version" : 1}} in collection: entitlementJobExecution
```

So this problem concern only MongoRepository

christophstrobl:
Thank you for reporting - we'll look into it.

christophstrobl:
[...]

PR #4921, opened 2025-03-11
Fix basic update overriding values.

christophstrobl:
This change makes sure `BasicUpdate` appends values to operations instead of overriding them. This change aligns the behaviour with `Update` and fixes issues where using the `@Update` annotation with versioned entities can lead to loss of update information.

Closes: #4918


mp911de:

Issue #4937, opened 2025-04-07
MongoClientVersion uses a non-reliable way to fetch db driver version.


smagellan:
Hello.

MongoClientVersion does its best to determine mongodb driver version properly and use a proper driver API. 
But looks like com.mongodb.internal.build.MongoDriverVersion.VERSION value is inlined and always 5.2.1 (for spring-data-mongodb 4.4.4), note LDC "5.2.1":
<details>
  <summary>Bytecode</summary>

```
private static getVersionFromPackage(Ljava/lang/ClassLoader;)Lorg/springframework/data/util/Version;
    // parameter  classLoader
  @Lorg/springframework/lang/Nullable;()
    TRYCATCHBLOCK L0 L1 L2 java/lang/IllegalArgumentException
   L3
    LINENUMBER 97 L3
    LDC "com.mongodb.internal.build.MongoDriverVersion"
    ALOAD 0
    INVOKESTATIC org/springframework/util/ClassUtils.isPresent (Ljava/lang/String;Ljava/lang/ClassLoader;)Z
    IFEQ L4
   L0
    LINENUMBER 99 L0
    LDC "5.2.1"
    INVOKESTATIC org/springframework/data/util/Version.parse (Ljava/lang/String;)Lorg/springframework/data/util/Version;
   L1
    ARETURN
   L2
    LINENUMBER 100 L2
   FRAME SAME1 java/lang/IllegalArgumentException
    ASTORE 1
   L4
    LINENUMBER 104 L4
   FRAME SAME
    ACONST_NULL
    ARETURN
   L5
    LOCALVARIABLE classLoader Ljava/lang/ClassLoader; L3 L5 0
    MAXSTACK = 2
    MAXLOCALS = 2
```
</details>
 This will not work properly if one e.g. excludes driver 5.X version and provides 4.X explicitly instead, see stacktrace:

<details>
  <summary>Stacktrace</summary>

```
Exception in thread "main" java.lang.ExceptionInInitializerError
	at org.springframework.data.mongodb.core.MongoTemplate.lambda$getCollectionNames$29(MongoTemplate.java:2353)
	at org.springframework.data.mongodb.core.MongoTemplate.execute(MongoTemplate.java:582)
	at org.springframework.data.mongodb.core.MongoTemplate.getCollectionNames(MongoTemplate.java:2351)
	at smagellan.test.mongodb.SpringDataMongodbRunner.main(SpringDataMongodbRunner.java:15)
[...]

PR #4938, opened 2025-04-07
Dynamically load com.mongodb.internal.build.MongoDriverVersion.VERSION

smagellan:
Dynamically load com.mongodb.internal.build.MongoDriverVersion.VERSION --
Closes gh-4937


mp911de:
I didn't notice your PR, but basically the fix for #4937 is doing that. In any case, thanks for looking into it.

Issue #4944, opened 2025-04-10
Inconsistent `_id` Storage Type Between `save` and `saveAll` for `@MongoId String id`


initdch:
### Description
When using Spring Data MongoDB with entities annotated with `@MongoId String id`, there is an inconsistency in how the `_id` field is stored in MongoDB depending on whether `save` or `saveAll` is used. Specifically, `save` stores the `_id` as a `String`, while `saveAll` stores it as an `ObjectId`, even though the ID is not explicitly set in either case.

That even leads to duplication in the DB, since when used `saveAll` and the `update` it will convert it to string but creates a duplicated object

### Steps to Reproduce
1. Set up a Spring Boot project with the following dependencies:
   - Spring Boot Starter Data MongoDB
2. Run the application and observe the output.
3. Check the DB
```
package com.example.mongo_bug;

import org.springframework.boot.CommandLineRunner;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.context.annotation.Bean;
import org.springframework.data.mongodb.core.mapping.Document;
import org.springframework.data.mongodb.core.mapping.MongoId;
import org.springframework.data.mongodb.repository.MongoRepository;

import java.util.List;

@SpringBootApplication
public class MongoBugApplication {

	public static void main(String[] args) {
		SpringApplication.run(MongoBugApplication.class, args);

	}

	@Bean
	public CommandLineRunner commandLineRunner(MyRepository repository) {
		return args -> {
			// Create entities without setting ID
			MyEntity entity1 = new MyEntity();
			entity1.setName("Entity1");
			MyEntity entity2 = new MyEntity();
			entity2.setName("Entity2");

			// Save using save
			repository.save(entity1);

			// Save using saveAll
			repository.saveAll(List.of(entity2));

			// Update the name of entity2 will lead to duplicate
			entity2.setName("UpdatedEntity2");
			repository.save(entity2);

			// Find all entities
			List<MyEntity> allEntities = repository.findAll();
			allEntities.forEach(entity -> System.out.println("Entity ID: " + entity.getId() + ", Name: " + entity.getName()));
		};
	}
}

// Repository interface
interface MyRepository extends MongoRepository<MyEntity, String> {
}

// Entity class
@Document
class MyEntity {
	@MongoId
	private String id;
	private String name;

	public String getId() {
		return id;
	}

	public void setId(String id) {
		this.id = id;
	}

[...]

PR #4945, opened 2025-04-11
Fix MongoId mapping for insertAll.

christophstrobl:
This PR fixes an issue where id properties annotated with `@MongoId` had not been converted into the desired target type when inserting a collection of objects instead a single one.

Closes: #4944


mp911de:
Todo: merge to `5.0.x` branch

PR #4949, opened 2025-04-17
Add support for fluent `QueryResultConverter`

mp911de:
We now support configuring a `QueryResultConverter` that has access to the raw result `Document` and the mapped result object that has been materialized from `MongoConverter`:

```java
List<Optional<Jedi>> result = template.query(Person.class)
                        .as(Jedi.class)
                        .matching(query(where("firstname").is("luke")))
                        .map((document, reader) -> Optional.of(reader.get()))
                        .all();
```

This is useful to either wrap, post-process results contextually or apply further transformations based on the result `Document`. Changes apply for imperative and reactive find, geoNear and aggregations.

TODO:

- [ ] `QueryResultConverter.entity()` (returning the converted object) is a good naming fit for queries, but with projections or aggregations we might to name it differently

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>4.5.0-SNAPSHOT</version>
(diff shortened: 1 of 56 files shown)


christophstrobl:
merged to _5.0.x_.

Issue #4920, opened 2025-03-11
Deprecate `MongoCustomConversions.BigDecimalRepresentation.STRING`


mp911de:
For Spring Data MongoDB 5.0, we should deprecate `MongoCustomConversions.BigDecimalRepresentation.STRING` (not for removal) and switch the `BigDecimal` default to `BigDecimalRepresentation.DECIMAL128` to indicate that `Decimal128` is the preferred data type enabling range queries.

Storing big decimals as string is still a valid option, deprecation expresses that there are certain limitations and applications can use that option if they are aware of the implications.

Follow-up to #3444 and #4916

PR #4957, opened 2025-04-22
Deprecate `MongoCustomConversions.BigDecimalRepresentation.STRING`

panic08:
It is better to use BigInteger storage as Decimal128, so we should deprecate storage as strings.

Fix: #4920


christophstrobl:
Thank you for taking the time to look into this. Since it is a behaviour breaking change there are several test failures that need to be addressed before we can bring the change in.

mp911de:

PR #4959, opened 2025-05-02
Fix AddFieldsOperationBuilder to treat String value as Field reference

kssumin:
String values in MongoDB aggregation operations should be treated as field references, but the current implementation in `AddFieldsOperationBuilder` produces invalid MongoDB syntax. MongoDB expects field references in the format `$fieldName`, but the current code generates an incompatible complex object structure.


christophstrobl:
Thank you for the PR @kssumin !

christophstrobl:
Thank you for your contribution - merged to _main_ development line and back ported to maintenance branches.

PR #4960, opened 2025-05-05
Explore Search Results

mp911de:
Return `SearchResult` from Repository Query methods.

Depends on https://github.com/spring-projects/spring-data-commons/pull/3285

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/VectorSearch.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/VectorSearch.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2025 the original author or authors.
+ * Copyright 2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,7 +41,7 @@ import org.springframework.data.mongodb.core.aggregation.VectorSearchOperation;
  *
  * @author Mark Paluch
  * @since 5.0
- * @see org.springframework.data.geo.Distance
+ * @see org.springframework.data.domain.Score
  * @see org.springframework.data.domain.Vector
  * @see org.springframework.data.domain.SearchResults
  */
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/VectorSearch.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/VectorSearch.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2016-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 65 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/VectorSearch.java
```diff
@@ -0,0 +1,119 @@
...
+@Query
+@Hint
+public @interface VectorSearch {
+
+	/**
+	 * Configuration whether to use ANN or ENN for the search. ANN is the default.
+	 *
+	 * @return the search type to use.
+	 */
+	VectorSearchOperation.SearchType searchType() default VectorSearchOperation.SearchType.ENN;
```
christophstrobl, line 60: Documentation (default: ANN) contradicted by implementation (default: ENN).
MongoDB itself will use `ANN` by default (so the doc) is closer to what happens than the actual default value.

I think the MongoDB restrictions here do not allow much flexibility regarding defaulting of values. Like on the one hand `ANN` ist stated as default but requires `numCandidates` while the `exact` flag is labelled optional.  

We could argue `DEFAULT` is a viable option here. Checking the `numCandidates` attribute, using `ENN` if no value present, `ANN` otherwise.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/MongoQueryExecution.java
```diff
@@ -210,6 +221,83 @@ private static boolean isListOfGeoResult(TypeInformation<?> returnType) {
 		}
 	}
 
+	/**
+	 * {@link MongoQueryExecution} to execute vector search.
+	 *
+	 * @author Mark Paluch
+	 * @since 5.0
+	 */
+	class VectorSearchExecution implements MongoQueryExecution {
```
christophstrobl, line 230: curious if we can derive a more AOT friendly variant of this.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/VectorSearchDelegate.java
```diff
@@ -0,0 +1,397 @@
...
+
+			Vector vector = accessor.getVector();
+			Score score = accessor.getScore();
+			Range<Score> distance = accessor.getScoreRange();
+			int limit;
+
+			if (query.isLimited()) {
+				limit = query.getLimit();
+			} else {
+				limit = Math.max(1, numCandidates() != null ? numCandidates() / 20 : 1);
```
christophstrobl, line 218: limit cannot exceed `numCandidates` and `numCandidates` should be at least 20x higher than limit but is deriving the limit the way it's done here valid?

mp911de: Probably doesn't make sense.

christophstrobl:
merged to _5.0.x_

PR #4964, opened 2025-05-07
Adopt to AOT changes in Commons

mp911de:
See https://github.com/spring-projects/spring-data-commons/pull/3282

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
@@ -15,15 +15,13 @@
  */
 package org.springframework.data.mongodb.repository.support;
 
-import static org.springframework.data.querydsl.QuerydslUtils.*;
-
 import java.io.Serializable;
 import java.lang.reflect.Method;
 import java.util.Optional;
 
 import org.jspecify.annotations.Nullable;
+
 import org.springframework.beans.factory.BeanFactory;
-import org.springframework.dao.InvalidDataAccessApiUsageException;
 import org.springframework.data.mapping.context.MappingContext;
 import org.springframework.data.mongodb.core.MongoOperations;
 import org.springframework.data.mongodb.core.mapping.MongoPersistentEntity;
@@ -35,7 +33,6 @@ import org.springframework.data.mongodb.repository.query.PartTreeMongoQuery;
 import org.springframework.data.mongodb.repository.query.StringBasedAggregation;
 import org.springframework.data.mongodb.repository.query.StringBasedMongoQuery;
 import org.springframework.data.projection.ProjectionFactory;
-import org.springframework.data.querydsl.QuerydslPredicateExecutor;
 import org.springframework.data.repository.core.NamedQueries;
 import org.springframework.data.repository.core.RepositoryInformation;
 import org.springframework.data.repository.core.RepositoryMetadata;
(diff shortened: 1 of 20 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/QuerydslContributor.java
```diff
@@ -0,0 +1,70 @@
...
+ * MongoDB-specific {@link RepositoryFragmentsContributor} contributing Querydsl fragments if a repository implements
+ * {@link QuerydslPredicateExecutor}.
+ *
+ * @author Mark Paluch
+ * @since 5.0
+ * @see QuerydslMongoPredicateExecutor
+ */
+enum QuerydslContributor implements MongoRepositoryFragmentsContributor {
+
+	INSTANCE;
```
onobc, line 38: This is an interesting way to create a singleton, that is ultimately what this is, correct? I like it, but it was not clear to me at first.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
```diff
@@ -60,6 +57,7 @@ public class MongoRepositoryFactory extends RepositoryFactorySupport {
 	private final CrudMethodMetadataPostProcessor crudMethodMetadataPostProcessor = new CrudMethodMetadataPostProcessor();
 	private final MongoOperations operations;
 	private final MappingContext<? extends MongoPersistentEntity<?>, MongoPersistentProperty> mappingContext;
+	private MongoRepositoryFragmentsContributor fragmentsContributor = MongoRepositoryFragmentsContributor.DEFAULT;
```
onobc, line 60: Both, the `MongoRepositoryFactory` and `MongoRepositoryFactoryBean` have a reference to `MongoRepositoryFragmentsContributor` and the latter sets the fragments contributor always. This makes me wonder if we should set the default instance in the factory since its always set by the factory bean?

onobc (commented):
The changes make sense to me. I will adapt what I have done in Spring Data Cassandra to fit this as well. I have a couple of comments but other than that LGTM.

christophstrobl:
The interface resolution when generating the `<repo>.json` file seems to be a little off for a repo interface that uses methods delegated to the default repository implementation when not using one of the predefined repository interfaces.
```java
interface OwnerRepository extends Repository<Owner, Long> {
  void save(Owner owner);
}
```

```json
{
  "name": "save",
  "signature": "public abstract void org.springframework.samples.petclinic.owner.OwnerRepository.save(org.springframework.samples.petclinic.owner.Owner)",
  "fragment": {
    "interface": "org.springframework.data.mongodb.repository.support.SimpleMongoRepository",
    "fragment": "org.springframework.data.mongodb.repository.support.SimpleMongoRepository"
  }
},
```

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/QuerydslContributor.java
mp911de: We use enums for selected singleton cases. Especially those stateless variants are good candidates.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
mp911de: `MongoRepositoryFactory` can be used standalone, without a factory bean. Introducing the `FragmentsContributor` concept comes from the need to know ahead of time, which fragments would be contributed. We cannot instantiate during AOT the `MongoRepositoryFactory` as it requires infrastructure (driver) access so we have to fallback to something else (factory bean).

mp911de:
Yeah, I have no good idea to fix that yet. The `interface` in fragments refers to the type that contributes signatures when looking up methods. For Querydsl and fragments, that is the actual interface. Maybe we just suppress `interface` if the fragment is the same as `interface`.

spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/MongoRepositoryFactory.java
onobc: Ahh got it. Makes sense to me now. Thanks for the explanation.

christophstrobl:
merged to _5.0.x_

PR #4971, opened 2025-05-14
refactor: Loggers should be named for their enclosing classes

jkschneider:
Hope this helps :)

Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.java.logging.slf4j.LoggersNamedForEnclosingClass?organizationId=U3ByaW5n


mp911de:
Thanks for reaching out. In that specific case we're obtaining the Logger from a different component to capture its logs. The logger declared in the test isn't to log something from the tests as one would typically assume.

It would generally make sense to run tests prior submitting pull requests.

PR #4977, opened 2025-05-19
Polishing.

mp911de:
Fix contract annotations, add missing Override annotations, make fields final where possible.


christophstrobl:
Thanks @mp911de - merged to _main_ development line.

Issue #5004, opened 2025-06-18
Add support for geospatial AOT-generated query methods


christophstrobl:
Add support for ahead of time code generation for query methods (`near` & `within`) operating upon geospatial data (may require special geo index to be present).

```java
List<Entity> findByLocationNear(Point point);

List<Entity> findByLocationNear(Point point, Distance maxDistance);

List<Entity> findByLocationWithin(Circle circle);

List<Entity> findByLocationWithin(Box box);

List<Entity> findByLocationWithin(Polygon polygon);

GeoResults<Entity> findByLocationNear(Point point, Distance maxDistance);

GeoResults<Entity> findByLocationNearAndLastname(Point point, Distance maxDistance, String Lastname);

GeoResults<Entity> findPersonByLocationNear(Point point, Range<Distance> distance);

GeoPage<Entity> findByLocationNear(Point point, Distance maxDistance, Pageable pageable);
```

PR #5005, opened 2025-06-18
Add support for AOT-generated geospatial/vector search queries, expression parameter bindings, and regular expressions

christophstrobl:
Closes: #5004
Closes: #5006

Also adds:
- general expression evaluation
- Collation support for queries


christophstrobl:
we should merge #5014 before this one and make sure to apply the `EnableIfVectorSearchAvailable` annotation.

PR #5020, opened 2025-07-17
Provide fallback for Observation KeyValues

jonatan-ivanov:
Similarly to https://github.com/spring-projects/spring-data-mongodb/pull/4994 this PR adds a fallback value to all the `KeyValue`s used by `DefaultMongoHandlerObservationConvention`.

TODO:
- [ ] Read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc)
- [ ] Use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide)
- [ ] Add test cases
- [ ] Added myself as author in the headers of the classes I touched

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoObservation.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoObservation.java
@@ -15,8 +15,11 @@
  */
 package org.springframework.data.mongodb.observability;
 
+import io.micrometer.common.KeyValue;
 import io.micrometer.common.docs.KeyName;
 import io.micrometer.observation.docs.ObservationDocumentation;
+import org.jspecify.annotations.Nullable;
+import org.springframework.util.ObjectUtils;
 
 /**
  * A MongoDB-based {@link io.micrometer.observation.Observation}.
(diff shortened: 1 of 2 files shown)


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/observability/MongoObservation.java
```diff
@@ -172,6 +175,15 @@ public String asString() {
...
 			}
+		};
+
+		/**
+		 * Creates a key value for the given key name.
+		 * @param value value for key, if value is null or empty {@link KeyValue.NONE_VALUE} will be used
+		 * @return key value
+		 */
+		public KeyValue withOptionalValue(@Nullable String value) {
+			return withValue(ObjectUtils.isEmpty(value) ? KeyValue.NONE_VALUE : value);
```
shakuzen, line 186: I wonder if "UNKNOWN" is a better default value for most of these rather than none.

jonatan-ivanov: @shakuzen 
>I wonder if "UNKNOWN" is a better default value for most of these rather than none.

I think so, I can also make the fallback mandatory so it needs to be decided for every key-value.

@mp911de
>I would like to have a bit more encapsulated API for a more declarative approach [...]

We can consider improving on this in Micrometer 1.16 (let me talk to @shakuzen about it) but I think this should be considered as a bug and should be fixed in earlier versions where Micrometer 1.16 will not be available I think (unless you want to copy some code from there).

jonatan-ivanov: https://github.com/micrometer-metrics/micrometer/issues/6564

Issue #5036, opened 2025-08-05
Default behavior for BigDecimal changed and broke custom converters


zeapo:
Since the commit https://github.com/spring-projects/spring-data-mongodb/commit/2b6730d1da07de07fea77a1939f986666819a7aa any WritingConverters for BigDecimal and BigInteger is no longer taken in priority; 

I understand that we can use the previous behavior by changing the BigDecimalRepresentation to DECIMAL128; however this is an explicit value that we have to set, otherwise any previously defined converter is no longer a priority. Which is a **breaking change**, as the upgrade documentation does not mention is new behavior.

Spring Data Commons take the first candidate that matches the source type if no explicit target type is defined. However, since the afore mentioned commit, Spring Data MongoDB registers its converters within the users' converters:

```java
if (bigDecimals == BigDecimalRepresentation.STRING) {
	converters.add(BigDecimalToStringConverter.INSTANCE);
	converters.add(StringToBigDecimalConverter.INSTANCE);
	converters.add(BigIntegerToStringConverter.INSTANCE);
	converters.add(StringToBigIntegerConverter.INSTANCE);
}

if (!useNativeDriverJavaTimeCodecs) {

	converters.addAll(customConverters);
	return new ConverterConfiguration(STORE_CONVERSIONS, converters, convertiblePair -> true,
			this.propertyValueConversions);
}
```
The behavior should be similar to what is done for the datetimes, the code should look something like (from my limited understanding of the way this works):

```java
private static final Set<Class<?>> BIG_DECIMAL_SIMPLE_TYPES = Set.of(BigDecimal.class, BigInteger.class);
....

if (!useNativeDriverJavaTimeCodecs) {
	StoreConversions storeConversions = StoreConversions
			.of(new SimpleTypeHolder(BIG_DECIMAL_SIMPLE_TYPES, MongoSimpleTypes.HOLDER), converters);

	return new ConverterConfiguration(storeConversions, this.customConverters, convertiblePair -> true,
			this.propertyValueConversions);
}
```

zeapo:
We don't mind that the updated default behavior be updated; the issue is that the change makes it ignore any previously defined converter for the BigDecimal;

Given this existing code:

```kotlin
  @WritingConverter
  class BigIntegerWriteConverter : Converter<BigInteger, Number>
....
  @Bean
  fun customConversions() =
      MongoCustomConversions.create {
          it.registerConverters(
              listOf(
                  BigIntegerWriteConverter(),
[...]

PR #5052, opened 2025-09-09
Fix regression in custom conversion ordering.

christophstrobl:
This PR fixes a regression that prevented user provided custom converters from taking precedence over framework defined ones.

Resolves: #5036


spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MongoCustomConversions.java
```diff
@@ -378,7 +378,7 @@ ConverterConfiguration createConverterConfiguration() {
 
 			if (!useNativeDriverJavaTimeCodecs) {
 
-				converters.addAll(customConverters);
+				converters.addAll(0, customConverters); // need to add at index 0!!
```
mp911de, line 381: Actually, shouldn't `converters` go into `STORE_CONVERSIONS`?

Issue #1861, opened 2014-06-24
Generally review logging handling [DATAMONGO-961]


spring-projects-issues:
**[Thomas Darimont](https://jira.spring.io/secure/ViewProfile.jspa?name=thomasd)** opened **[DATAMONGO-961](https://jira.spring.io/browse/DATAMONGO-961?redirect=false)** and commented

Currently MongoDbUtils uses the LOGGER directly without using any guards to check whether the logging is enabled

---

**Affects:** 1.5 GA (Dijkstra)

**Referenced from:** pull request https://github.com/spring-projects/spring-data-mongodb/pull/194

PR #5188, opened 2026-04-26
Render MongoTemplate ObjectId logs in shell format

### LINKED ISSUE #4873: Queries in MongoTemplate Logs with ObjectId Format.
When any operations made through MongoTemplate, I set the log level to DEBUG using the following configuration:
logging.level.org.springframework.data.mongodb.core.MongoTemplate=DEBUG.

This allows me to view the queries executed by MongoTemplate, which is helpful. However, the issue is that the query logs display ObjectId in the oid format, which cannot be directly executed in the MongoDB Compass client.

Is there a way to log the queries in the ObjectId("") format instead? I couldn’t find a solution for this issue. If a solution already exists, please let me know. If not, it would be great to have this functionality implemented.

codingkiddo:
Closes #4873

This change updates MongoTemplate debug query logging to render BSON ObjectId values using Mongo shell-style JSON formatting.

Before:

```json
{ "_id" : { "$oid" : "507f1f77bcf86cd799439011" } }

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java
@@ -136,6 +136,9 @@ import com.mongodb.client.model.*;
 import com.mongodb.client.result.DeleteResult;
 import com.mongodb.client.result.UpdateResult;
 
+import org.bson.json.JsonMode;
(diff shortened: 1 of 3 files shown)


mp911de:
Logging is a larger theme discussed in #4873 and #1861 requiring a wider consideration of how to allow configuration to help with debugging and running queries from log outputs. Closing as duplicate.

PR #5193, opened 2026-05-05
Fix GeoJson polygon conversion for multiple inner rings

### LINKED ISSUE #5190: GeoConverters does only support polygons with one hole?
Hi all,

I currently try to read a `GeoJsonPolygon` with two holes from the database. In the db its correctly persisted with three items in the coordinates array (one outer and two inner rings). 

However, when reading, it seems that only the first hole is considered. This issue was already discussed in https://github.com/spring-projects/spring-data-mongodb/issues/4104 with the same remark in the end. I am not sure if it was not implemented for a specific reason or missed?

Kind regards
Chris

DragonFSKY:
Closes #5190

This PR updates `GeoConverters.toGeoJsonPolygon(...)` to preserve all inner rings when reading a `GeoJsonPolygon` from a MongoDB `Document`.

The existing behavior for polygons without inner rings and with a single inner ring remains unchanged. Additional coordinate rings after the outer ring are now applied as inner rings.

Tests cover both direct `GeoJsonPolygon` conversion and the `GeoJsonMultiPolygon` path, since it reuses the same conversion method.

Tests:
- `./mvnw -s settings.xml -pl spring-data-mongodb clean test '-Dtest=GeoJsonConverterUnitTests$DocumentToGeoJsonPolygonConverterUnitTests,GeoJsonConverterUnitTests$DocumentToGeoJsonMultiPolygonConverterUnitTests'`

Checklist:
- [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Do not submit any formatting related changes.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/GeoConverters.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/GeoConverters.java
@@ -60,6 +60,7 @@ import com.mongodb.Function;
  * @author Oliver Gierke
  * @author Christoph Strobl
  * @author Thiago Diniz da Silveira
+ * @author dragonfsky
  * @since 1.5
  */
 @SuppressWarnings("ConstantConditions")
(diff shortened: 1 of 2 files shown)


christophstrobl:
Thank you @DragonFSKY for picking up the issue. We're already in the post RC phase and will get back to this PR in the next iteration.

PR #5194, opened 2026-05-05
Support array-backed Point mapping

### LINKED ISSUE #4997: Mapping should be able to read Point coordinates in array format.
The mapping layer should allow to reading back coordinates stored in raw array format `[-73.99171, 40.738868]` into a `Point` type.

DragonFSKY:
Closes #4997

This PR adds opt-in support for array-backed `Point` mapping through `@Field(targetType = FieldType.ARRAY)`.

With the annotation, coordinate arrays such as `[-73.99171, 40.738868]` are read into `Point` and written back as coordinate arrays. Query and update mapping for the same property also keeps the array representation.

The default `Point` mapping remains unchanged and continues to use the existing `{ x, y }` document representation.

Tests:
- `./mvnw -pl spring-data-mongodb -Dtest=GeoConvertersUnitTests,MappingMongoConverterUnitTests,QueryMapperUnitTests,UpdateMapperUnitTests -Dmaven.compiler.proc=none test`
- `git diff --check`

Checklist:
- [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Do not submit any formatting related changes.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/GeoConverters.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/GeoConverters.java
@@ -45,7 +45,10 @@ import org.springframework.data.mongodb.core.geo.GeoJsonMultiPolygon;
 import org.springframework.data.mongodb.core.geo.GeoJsonPoint;
 import org.springframework.data.mongodb.core.geo.GeoJsonPolygon;
 import org.springframework.data.mongodb.core.geo.Sphere;
+import org.springframework.data.mongodb.core.mapping.FieldType;
+import org.springframework.data.mongodb.core.mapping.MongoPersistentProperty;
 import org.springframework.data.mongodb.core.query.GeoCommand;
+import org.springframework.data.mongodb.util.BsonUtils;
 import org.springframework.lang.Contract;
 import org.springframework.util.Assert;
 import org.springframework.util.NumberUtils;
@@ -60,6 +63,7 @@ import com.mongodb.Function;
  * @author Oliver Gierke
  * @author Christoph Strobl
  * @author Thiago Diniz da Silveira
+ * @author dragonfsky
  * @since 1.5
  */
 @SuppressWarnings("ConstantConditions")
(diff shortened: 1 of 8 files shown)


christophstrobl:
Thank you @DragonFSKY for providing the PR. Since we already released an RC for the current release train we might consider your contribution for the next iteration.

Until we have time to do a full review of the proposed changes I'd like to leave a comment already that I think needs to be addressed.

Currently it is not possible to auto convert `coordinates : [x,y]` into a `Point coordinates` field. 
The proposed changes fix this, but would also, when saving back the entity, silently change the stored document to `coordinates : { x: ..., y: ...}`.

I could think of something like the following (which is currently not supported right now) to prevent this from happening.
```java
@Field(targetType = FieldType.ARRAY)
Point coordinates;
``` 

Please, the above is just from a quick glance and should not be taken as implementation advice. Let's discuss things before spending more time changing the PR.

DragonFSKY:
Sorry, my initial thinking was incomplete here. I focused on the read side first and did not think through the read/write storage shape carefully enough.

Looking at the current upstream mapping code, `Point` is still written through the existing `{ x, y }` representation. So if array coordinates are accepted as `Point` on read, the storage shape probably needs to be considered explicitly as well; otherwise a later write could store a different representation.

My current thought is that array-backed `Point` mapping might need to be opt-in, for example:

```java
class Location {

    @Field(targetType = FieldType.ARRAY)
    Point coordinates;
}
```

This way, an explicitly array-backed property could preserve the array representation:

```javascript
{ "coordinates": [ -73.99171, 40.738868 ] }
```

while existing `Point` mappings without the annotation would keep using the current default representation:

```javascript
{ "coordinates": { "x": -73.99171, "y": 40.738868 } }
```

Does that sound like a reasonable direction? If you see a better approach, please feel free to point me in that direction.

DragonFSKY:
I updated the PR after my previous comment to make array-backed `Point` mapping explicit via `@Field(targetType = FieldType.ARRAY)`.

With that annotation, reads, writes, queries, and updates use the `[x, y]` representation. The default `Point` mapping remains unchanged and still writes as `{ x, y }`, so existing documents should not silently change shape on save.

If there are any remaining concerns or preferred adjustments, I will follow up.

PR #5218, opened 2026-06-19
Honor standalone @Collation on entities and propagate to repository base methods

### LINKED ISSUE #4535: @Collation annotation on documents, repositories or queries not interpreted
I'm using Spring data mongo and want to use collations for various queries. The collations need to work with the default repository methods as well as with custom queries. 

According to the [reference doc](https://docs.spring.io/spring-data/mongodb/reference/mongodb/repositories/query-methods.html#mongo.repositories.collation) and the #2755 , the following should work: 

```kotlin
@Collation(value = "...")
data class SampleDocument(
    @field:MongoId
    var id: String? = null,
    val value: String
)

interface RepositoryWithCollationInQuery: CrudRepository<SampleDocument, String> {
    @Collation(value = "...")
    @Query(
        value = "{}",
        sort = """{ "value":  1}""",
    )
    fun findAllCollationInAnnotation(): List<SampleDocument>
}
```

But the collation is not evaluated. It is only evaluated when adding it to `@Query` directly.

A sample project with the expected code (and failing tests) can be found here: https://github.com/dirkbolte/mongodb_collation_test

The following approaches were tried:

- adding `@Collation` to the data class (failed)
- adding `@Collation` to the repository interface (failed)
- adding `@Collation` to the query method (failed)
- specifying `collation=...` on the `@Query` annotation (worked)

I would expect that adding the `@Collation` annotation to repository and/or document would allow all repository methods to make use of it, especially when using `findAll(pageable)` of the default spring repository interfaces. When requiring `@Query` for everything, I would have to specify every query manually and explicitly, which doesn't feel right.

Versions:
- Spring Boot: 3.1.5
- Spring Data Mongo: 4.1.5
- MongoDB: 6 (4 showed no differences)

For reference, the I asked the question on [stack overflow](https://stackoverflow.com/questions/77325494/using-collations-with-spring-data-mongo-seems-not-to-work) as well, in case this is a user error.

seonwooj0810:
## What this addresses

`@Collation` on an entity is not picked up unless it is set through `@Document(collation = ...)`, and even when it _is_ picked up, the entity collation only reaches `Example`-based repository methods. Plain CRUD methods (`findById`, `findAll`, `count`, `deleteAll`, …) are dispatched without it, so the resulting MongoDB operation runs with no collation even though the entity declares one.

Closes: #4535

## Root cause

Two cooperating gaps in the entity → repository wiring:

1. `BasicMongoPersistentEntity` only consults `@Document.collation()`. The standalone `@Collation` annotation in `org.springframework.data.mongodb.core.annotation` is the canonical source (it is what `@Document.collation()` is itself `@AliasFor` for), yet it is ignored unless `@Document` is present and uses the `collation` attribute. So `@Collation(\"en_US\")` on a type — with or without an accompanying `@Document` — silently has no effect.

2. `SimpleMongoRepository` / `SimpleReactiveMongoRepository` only propagate `entityInformation.getCollation()` on the `Example`-based query path. The basic CRUD methods (`findById`, `existsById`, `findAll`, `findAll(Sort)`, `findAll(Pageable)`, `findAllById`, `count`, `deleteById`, `deleteAllById`, `deleteAll`) build their `Query` without `.collation(...)`, so even if (1) is fixed, the collation is dropped at the repository boundary.

These are exactly the two locations called out in the issue thread.

## Change summary

**`BasicMongoPersistentEntity`** — resolve collation via `AnnotatedElementUtils.findMergedAnnotation(rawType, Collation.class)`. Because `@Document` is meta-annotated with `@Collation` and `Document#collation` is declared as `@AliasFor(annotation = Collation.class, attribute = \"value\")`, the merged lookup returns the same value whether the user writes `@Document(collation = \"en_US\")` or `@Collation(\"en_US\")` (with or without `@Document`). Existing `@Document.collation()` behavior — literal value, SpEL expression, JSON document form — is preserved.

**`SimpleMongoRepository`** and **`SimpleReactiveMongoRepository`** — add a small `applyDefaultCollation(Query)` helper and invoke it from every base CRUD method that builds a `Query` (and from the shared `findAll(Query)` / `getIdQuery(Iterable)` helpers that funnel `findAll` and `findAllById`). The helper only sets the collation when one is defined and the query does not already carry one, so an explicit `Query.collation(...)` from a caller still wins.

## Why this is the right fix

The maintainer (christophstrobl) diagnosed exactly this shape on the issue: \"`BasicPersistentEntity` did not properly consider `@Collation` but was only looking for `@Document` and the collation was not properly passed on in `SimpleMongoRepository`.\" The change is additive — annotation handling extends to a strictly larger set of users (the existing `@Document(collation=...)` users keep getting the same `Collation` instance), and collation propagation aligns the base CRUD methods with the `Example`-based methods that have always carried it. Behavior for users that do not declare an entity collation is unchanged: `entityInformation.hasCollation()` returns `false` and `applyDefaultCollation` is a no-op.

## Tests

New tests added:

- `BasicMongoPersistentEntityUnitTests`:
  - `readsStandaloneCollationAnnotation` — `@Document` + standalone `@Collation(\"en_US\")` resolves to `Collation.of(\"en_US\")`.
  - `readsStandaloneCollationAnnotationWithoutDocumentAnnotation` — `@Collation(\"en_US\")` without `@Document` resolves correctly.
- `SimpleMongoRepositoryUnitTests`:
  - `shouldAddDefaultCollationToFindMethods` (parameterized over `findAll()`, `findAll(Sort)`, `findAll(Pageable)`).
  - `shouldAddDefaultCollationToFindById`, `shouldAddDefaultCollationToExistsById`, `shouldAddDefaultCollationToCount`, `shouldAddDefaultCollationToDeleteById`, `shouldAddDefaultCollationToDeleteAll`.
- `SimpleReactiveMongoRepositoryUnitTests`:
  - `shouldAddDefaultCollationToFindById`, `shouldAddDefaultCollationToCount`, `shouldAddDefaultCollationToDeleteAll`.

Existing collation tests (`DATAMONGO-1854`, `DATAMONGO-2565`) continue to pass; they cover `@Document(collation = ...)` literal/JSON/SpEL forms.

## Verification done

- ✅ Reviewed the issue thread and confirmed the maintainer-pinpointed root cause matches the code state on `main` (BasicMongoPersistentEntity ignores standalone `@Collation`; SimpleMongoRepository drops collation on base CRUD methods).
- ✅ Searched `gh pr list` and PR search on `spring-projects/spring-data-mongodb` for any in-flight PR referencing #4535 — none.
- ✅ Confirmed no \"I'm working on this\" claim from another contributor in the last 21 days (last comments are user bumps from 2025-03 and 2025-07-19).
- ✅ Grepped `main` for `BasicMongoPersistentEntity`, `SimpleMongoRepository`, `SimpleReactiveMongoRepository`, and `entityInformation.getCollation()` to confirm the bug pattern is still present.
- ✅ Targeted Maven build: `./mvnw -pl spring-data-mongodb test -Dtest='BasicMongoPersistentEntityUnitTests,SimpleMongoRepositoryUnitTests,SimpleReactiveMongoRepositoryUnitTests'` → 61 tests, all green.
- ✅ Broader regression: `./mvnw -pl spring-data-mongodb test -Dtest='BasicMongoPersistent*,*Mongo*Repository*UnitTests,*MongoQuery*UnitTests'` → 365 tests, all green.

## Scope notes

The issue also mentions `@Collation` on the repository interface and on individual query methods. Those flow through `MongoQueryMethod` (a separate code path from `BasicMongoPersistentEntity` / `SimpleMongoRepository`) and were intentionally left out to keep this change focused on the entity-level case the maintainer explicitly confirmed; method-level `@Collation` already works through the `@Query(collation = ...)` alias.

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/BasicMongoPersistentEntity.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/BasicMongoPersistentEntity.java
@@ -27,6 +27,7 @@ import java.util.Map;
 
 import org.jspecify.annotations.Nullable;
 
+import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.data.annotation.Id;
 import org.springframework.data.core.TypeInformation;
 import org.springframework.data.expression.ValueEvaluationContext;
@@ -96,13 +97,24 @@ public class BasicMongoPersistentEntity<T> extends BasicPersistentEntity<T, Mong
 			this.collection = StringUtils.hasText(document.collection()) ? document.collection() : fallback;
 			this.language = StringUtils.hasText(document.language()) ? document.language() : "";
 			this.expression = detectExpression(document.collection());
-			this.collation = document.collation();
-			this.collationExpression = detectExpression(document.collation());
 		} else {
 
 			this.collection = fallback;
 			this.language = "";
 			this.expression = null;
+		}
+
+		// Pick up the standalone @Collation annotation independently of @Document. @Document.collation
+		// remains supported because @Document is meta-annotated with @Collation and Document#collation
+		// is declared as @AliasFor(annotation = Collation.class, attribute = "value"), so the merged
+		// lookup returns the same value in either case.
+		org.springframework.data.mongodb.core.annotation.Collation collationAnnotation = AnnotatedElementUtils
+				.findMergedAnnotation(rawType, org.springframework.data.mongodb.core.annotation.Collation.class);
+
+		if (collationAnnotation != null && StringUtils.hasText(collationAnnotation.value())) {
+			this.collation = collationAnnotation.value();
+			this.collationExpression = detectExpression(collationAnnotation.value());
+		} else {
 			this.collation = null;
 			this.collationExpression = null;
 		}
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleMongoRepository.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/support/SimpleMongoRepository.java
@@ -124,6 +124,7 @@ public class SimpleMongoRepository<T, ID> implements MongoRepository<T, ID> {
 		Assert.notNull(id, "The given id must not be null");
 
 		Query query = getIdQuery(id);
+		applyDefaultCollation(query);
 		getReadPreference().ifPresent(query::withReadPreference);
 
 		return Optional.ofNullable(
@@ -136,6 +137,7 @@ public class SimpleMongoRepository<T, ID> implements MongoRepository<T, ID> {
 		Assert.notNull(id, "The given id must not be null");
 
 		Query query = getIdQuery(id);
+		applyDefaultCollation(query);
 		getReadPreference().ifPresent(query::withReadPreference);
 
 		return mongoOperations.exists(query, entityInformation.getJavaType(), entityInformation.getCollectionName());
@@ -158,6 +160,7 @@ public class SimpleMongoRepository<T, ID> implements MongoRepository<T, ID> {
 	public long count() {
 
 		Query query = new Query();
+		applyDefaultCollation(query);
 		getReadPreference().ifPresent(query::withReadPreference);
 		return mongoOperations.count(query, entityInformation.getCollectionName());
 	}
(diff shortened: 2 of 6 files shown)


seonwooj0810:
Gentle ping — open since mid-June.

Two gaps behind #4535: `BasicMongoPersistentEntity` only reads `@Document.collation()` and ignores the standalone `@Collation` annotation, and even when a collation is resolved it only reaches `Example`-based repository methods — plain CRUD (`findById`, `findAll`, `count`, `deleteAll`) dispatches without it. The PR closes both.

Is propagating the entity collation to the base CRUD methods a change you'd want in a minor, or would you rather scope this PR to the `@Collation` lookup only? Still merges cleanly, checks green.

seonwooj0810:
Following up: #4535 has been open since 2023, and with this change repository CRUD methods (`findById`, `findAll`, `count`, …) finally honor an entity's `@Collation` instead of silently running without it. Branch is conflict-free with green CI; happy to split the annotation lookup and the repository propagation into separate PRs if that eases review.

PR #5221, opened 2026-07-01
Improve SimpleMongoRepository.saveAll to use batch updates

### LINKED ISSUE #5220: Improve `MongoRepository.saveAll` to use batch updates
MongoDB supports efficient batch operations through bulkWrite, and `SimpleMongoRepository.saveAll(...)` already performs a batch insert when all entities are new . 

However, if entities contains any non-new entity, it calls `SimpleMongoRepository.save(...)` for each entity, which seems neither efficient nor consistent.

A possible improvement would be something like the following:

```java
List<S> newEntities = new ArrayList<>();
List<S> existingEntities = new ArrayList<>();

// ...

if (!newEntities.isEmpty()) {
   mongoOperations.insert(newEntities, entityInformation.getCollectionName());
}

if (!existingEntities.isEmpty()) {

   BulkOperations bulkOps = mongoOperations.bulkOps(BulkMode.ORDERED, entityInformation.getJavaType(),
            entityInformation.getCollectionName());

   for (S entity : existingEntities) {
      Query query = new Query(where(entityInformation.getIdAttribute()).is(entityInformation.getId(entity)));
      bulkOps.replaceOne(query, entity, FindAndReplaceOptions.options().upsert());
   }

   bulkOps.execute();
}
```

syjdev:

Closes #5220

### DIFF
--- a/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/BulkWriter.java
+++ b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/BulkWriter.java
@@ -47,6 +47,7 @@ import com.mongodb.client.model.bulk.ClientBulkWriteResult;
  * Internal API wrapping a {@link MongoTemplate} to encapsulate {@link Bulk} handling.
  *
  * @author Christoph Strobl
+ * @author Sangyeop Jeong
  * @since 5.1
  */
 class BulkWriter extends BulkWriterSupport {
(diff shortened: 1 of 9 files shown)


mp911de:
As per https://github.com/spring-projects/spring-data-mongodb/issues/5220#issuecomment-5089454996, we should seek a consistent bulk approach for inserts and updates, also considering the reactive implementation.

syjdev:
Thank you for the feedback. I'll revise the implementation based on both your comment on #5220 and the feedback you've provided here. Thanks!

syjdev:
I moved both the imperative and the reactive `saveAll(Iterable)` onto
`bulkWrite(Bulk, BulkWriteOptions)`.

Also, I changed `BulkWriter`/`ReactiveBulkWriter` to propagate the generated `_id`
back to the entity after the write because, unlike `doInsert(…)`, the bulk insert
path did not.

syjdev:
`saveAll(…)` now returns the instances produced by the write, so immutable entities come back with their generated `_id`.

This requires a new `BulkWriteResult.savedEntities()`, which throws `UnsupportedOperationException` for results that do not track entities. Let me know if you would prefer a different API shape.

PR #5226, opened 2026-07-21
Reverts change towards immutability

### LINKED ISSUE #5214: Possible regression in Criteria.and(String) behavior between Spring Data MongoDB 5.0.7 and 5.1.0
## Description

After upgrading from Spring Data MongoDB 5.0.7 to 5.1.0, we observed a behavior change when using `org.springframework.data.mongodb.core.query.Criteria`.

The following test passes with Spring Data MongoDB 5.0.7:

```java
class CriteriaTest {

    @Test
    void test() {
        final Criteria criteria = Criteria.where("ID1").is("1");
        criteria.and("ID2").is("2");

        Query query = Query.query(criteria);

        Assertions.assertEquals(
                "Query: { \"ID1\" : \"1\", \"ID2\" : \"2\"}, Fields: {}, Sort: {}",
                query.toString());
    }
}
```

The resulting query is:

```json
{ "ID1" : "1", "ID2" : "2" }
```

After upgrading to Spring Data MongoDB 5.1.0, the same test fails. The resulting query only contains the first criterion:

```json
{ "ID1" : "1" }
```

Changing the code to use the returned `Criteria` instance restores the expected behavior:

```java
Criteria criteria = Criteria.where("ID1").is("1");
criteria = criteria.and("ID2").is("2");
```

## Impact

The observed behavior suggests that `Criteria.and(String)` no longer updates the original `Criteria` instance and instead only applies changes to the returned object.

Existing code continues to compile successfully, but query semantics may change at runtime after upgrading. This can be difficult to detect, especially in applications without dedicated tests covering the generated MongoDB queries.

We were unable to find any mention of this behavior change in the Spring Data MongoDB 5.1.0 release notes and therefore wanted to report it as a possible regression.

## Environment

* Spring Boot 4.0.7 / Spring Data MongoDB 5.0.7
* Spring Boot 4.1.0 / Spring Data MongoDB 5.1.0
* Java: 25

schauder:
Criteria objects now properly mutate themselves as they used to.

Closes #5214

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-mongodb-parent</artifactId>
-	<version>5.2.0-SNAPSHOT</version>
+	<version>5.2.0-5214-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data MongoDB</name>
--- a/spring-data-mongodb-distribution/pom.xml
+++ b/spring-data-mongodb-distribution/pom.xml
@@ -13,7 +13,7 @@
 	<parent>
 		<groupId>org.springframework.data</groupId>
 		<artifactId>spring-data-mongodb-parent</artifactId>
-		<version>5.2.0-SNAPSHOT</version>
+		<version>5.2.0-5214-SNAPSHOT</version>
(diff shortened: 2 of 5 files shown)


schauder:
This is merged.

PR #5242, opened 2026-09-10
Remove redundant assertion in EncryptionOptions constructor

### LINKED ISSUE #5241: Redundant assertion in EncryptionOptions constructor
The three-argument `EncryptionOptions` constructor asserts `key` twice:

```java
public EncryptionOptions(String algorithm, EncryptionKey key,
        @Nullable QueryableEncryptionOptions queryableEncryptionOptions) {

    Assert.hasText(algorithm, "Algorithm must not be empty");
    Assert.notNull(key, "EncryptionKey must not be empty");
    Assert.notNull(key, "QueryableEncryptionOptions must not be empty");

    this.key = key;
    this.algorithm = algorithm;
    this.queryableEncryptionOptions = queryableEncryptionOptions;
}
```

The third assertion re-checks `key` while its message names `queryableEncryptionOptions`. That parameter is `@Nullable` and the two-argument constructor passes `null` for it, so it must not be validated. The assertion is therefore both redundant and misleading and should be removed.

kdomo:
The three-argument `EncryptionOptions` constructor asserts `key` twice:

```java
Assert.hasText(algorithm, "Algorithm must not be empty");
Assert.notNull(key, "EncryptionKey must not be empty");
Assert.notNull(key, "QueryableEncryptionOptions must not be empty");
```

The third assertion re-checks `key` while its message names `queryableEncryptionOptions`. That parameter is `@Nullable` and the two-argument constructor passes `null` for it, so it must not be validated. The assertion is both redundant and misleading, so this removes it.

Closes #5241


christophstrobl:
Thank you @kdomo - merged to `main` development line and back ported to `5.1.x` and `5.0.x`.

